PR #34, opened 2018-02-02
Started implementing use of constants for names of various fields and…

cmoesgaard:
… keys

job_title -> job_function
ou -> org_unit_type
address -> address_type

'Type' keys gjort mere specifikke. e.g.
type -> engagement_type


mora/service/facet.py
```diff
@@ -22,23 +22,24 @@
...
 blueprint = flask.Blueprint('facet', __name__, static_url_path='',
                             url_prefix='/service')
 
 FACETS = {
-    'address': 'Adressetype',
-    'job-function': 'Stillingsbetegnelse',
-    'ou': 'Enhedstype',
+    ADDRESS_TYPE: 'Adressetype',
+    JOB_FUNCTION: 'Stillingsbetegnelse',
+    ORG_UNIT_TYPE: 'Enhedstype',
```
danchr: We should use keys for the strings at the right as well. Also, note that you actually change the URLs to use underscore — is this intentional? (I'm not a fan of underscores in URLs — if so, we should consider converting them to query parameters.)

mora/service/role.py
```diff
@@ -17,31 +17,28 @@
...
+                     update_payload)
+from .keys import ORG, ORG_UNIT, ROLE_TYPE, VALID_FROM, VALID_TO
+from .mapping import (ORG_FUNK_GYLDIGHED_FIELD,
+                      ORG_FUNK_TYPE_FIELD, ORG_UNIT_FIELD,
+                      ROLE_FIELDS)
 
 blueprint = flask.Blueprint('roles', __name__, static_url_path='',
                             url_prefix='/service')
 
 ROLE_KEY = 'Rolle'
```
danchr: Ditto.

mora/service/engagement.py
```diff
@@ -18,25 +18,22 @@
...
+from .keys import (ENGAGEMENT_TYPE, JOB_FUNCTION, ORG, ORG_UNIT, PERSON,
+                   VALID_FROM, VALID_TO)
+from .mapping import (ENGAGEMENT_FIELDS, JOB_FUNCTION_FIELD,
+                      ORG_FUNK_GYLDIGHED_FIELD, ORG_FUNK_TYPE_FIELD,
+                      ORG_UNIT_FIELD)
 from .. import util
 
 blueprint = flask.Blueprint('engagements', __name__, static_url_path='',
                             url_prefix='/service')
 ENGAGEMENT_KEY = 'Engagement'
```
danchr: This should move to `keys`.

tests/test_integration_importing.py
```diff
@@ -1157,11 +1156,12 @@ def test_service_with_ballerup(self):
...
-                       'name': 'Ansat',
-                       'scope': None,
-                       'user_key': 'Ansat',
-                       'uuid': '39dd14ed-faa9-40bf-9fc9-13c440078458'},
+              'engagement_type': {
+                  'example': None,
+                  'name': 'Ansat',
+                  'scope': None,
+                  'user_key': 'Ansat',
+                  'uuid': '39dd14ed-faa9-40bf-9fc9-13c440078458'},
```
danchr: I assume we'll use a different key for associations?

mora/service/facet.py
cmoesgaard: My idea here, is to have the names be competely identical between getting them from the the facet endpoint to adding them as a parameter to a request. E.g. the facet from 'job_function' goes in the 'job_function' field - No confusion, no ambiguity.

mora/service/engagement.py
cmoesgaard: Done.

mora/service/role.py
cmoesgaard: Done.

mora/service/facet.py
danchr: We'll let them be for now; we can drop the underscore later, and perhaps drop it later when everything works.

danchr:
As agreed, looks good; just needs a rebase.

mora/service/engagement.py
```diff
@@ -21,28 +21,30 @@
...
 blueprint = flask.Blueprint('engagements', __name__, static_url_path='',
                             url_prefix='/service')
-ENGAGEMENT_KEY = 'Engagement'
 
-JOB_TITLE = 'job_title'
-ENGAGEMENT_TYPE = 'engagement_type'
-ORG_UNIT = 'org_unit'
-ORG = 'org'
+FUNCTION_KEYS = {
+    'engagement': ENGAGEMENT_KEY,
```
danchr: Should these belong to mapping as well?

cmoesgaard: Probably.

PR #35, opened 2018-02-05
IT systems, creating thereof

danchr:
This implements the “easy” parts of handling IT systems. Since we model addresses the same way — as a simple relation — we should probably let this get in. Editing will have to wait a bit.


mora/service/itsystem.py
```diff
@@ -15,18 +15,23 @@
...
+from .. import lora
 from .. import util
 
 from . import common
+from . import mapping
 
 blueprint = flask.Blueprint('itsystem', __name__, static_url_path='',
                             url_prefix='/service')
 
+ITSYSTEM = 'itsystem'
```
cmoesgaard: This should be in keys.py

mora/service/itsystem.py
```diff
@@ -182,3 +187,65 @@ def convert(start, end, effect):
...
+        return func(employee_uuid, req)
+
+    return wrapper
+
+
+@validate_it
+def create_system(employee_uuid, req):
+    systemid = req[ITSYSTEM].get('uuid')
+    valid_from = req.get('valid_from')
+    valid_to = req.get('valid_to') or 'infinity'
```
cmoesgaard: There's constants for valid_from and valid_to in keys.py - just for consistency.

PR #37, opened 2018-02-05
IT systems — part deaux

danchr:
Creation, writing and editing of IT systems. You might want to look at the individual commits rather than the whole blob.


danchr:
> The process of generating the 'original' without old_entry could probably be generalized a bit, to make this more reusable.
> Let's leave that for a future exercise for now.

Agreed; I plan to do generalisation as part of the address implementation.

PR #40, opened 2018-02-08
Implemented reading for Orlov and Rolle.

cmoesgaard:
Generalized existing reading code slightly to facilitate this.

### DIFF
--- a/mora/service/employee.py
+++ b/mora/service/employee.py
@@ -20,7 +20,7 @@ import flask
 from mora import lora
 from . import keys
 from . import common
-from .association import (ASSOCIATION_KEY, create_association,
+from .association import (create_association,
                           edit_association, terminate_association)
 from .engagement import (create_engagement, edit_engagement,
                          terminate_engagement)
@@ -607,7 +607,7 @@ def terminate_employee(employee_uuid):
 
     associations = c.organisationfunktion.get_all(
         tilknyttedebrugere=employee_uuid,
-        funktionsnavn=ASSOCIATION_KEY)
+        funktionsnavn=keys.ASSOCIATION_KEY)
     for association in associations:
         association_uuid = association[0]
         terminate_association(association_uuid, date)
--- a/mora/service/employee.py
+++ b/mora/service/employee.py
@@ -22,9 +22,8 @@ from . import keys
 from . import common
 from .association import (ASSOCIATION_KEY, create_association,
                           edit_association, terminate_association)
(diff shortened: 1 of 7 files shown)


mora/service/employee.py
```diff
@@ -22,9 +22,8 @@
 from . import common
 from .association import (ASSOCIATION_KEY, create_association,
```
danchr: Skal denne her også ind i `keys.py`?

mora/service/engagement.py
```diff
@@ -216,7 +211,10 @@ def get_employee_id(effect):
         return effect['relationer']['tilknyttedebrugere'][-1]['uuid']
 
     def get_unit_id(effect):
-        return effect['relationer']['tilknyttedeenheder'][-1]['uuid']
+        try:
+            return effect['relationer']['tilknyttedeenheder'][-1]['uuid']
+        except (KeyError, IndexError):
+            return None
```
danchr: Could you add a comment explaining which cases trigger this? It's not obvious to me.

mora/service/employee.py
cmoesgaard: Den ligger i øvrigt allerede i keys, dét der er bare en reference til referencen i keys.py.
Jeg fikser det.

mora/service/engagement.py
```diff
@@ -231,6 +229,70 @@ def get_title_id(effect):
...
+            }
+        }
+
+    converters = {
+        'engagement': convert_engagement,
+        'association': convert_association,
+        'role': convert_role,
+        'leave': convert_leave,
+    }
+
```
danchr: Looks fine for now, but I believe we can make it even more general — we'll eventually want to have this fully configurable, I believe.

The validity assignment seems redundant, though; perhaps you could move it to the caller?

mora/service/engagement.py
cmoesgaard: This is relevant for 'Leave' as we only have relations to users, not org units.

danchr: Precisely; it'd be nice to have a line noting that in the code.

mora/service/engagement.py
cmoesgaard: I agree - I thought about how to make this fully configurable and eventually expand on my rudimentary mappings and include them in this.
It's a somewhat non-trivial task, so I settled on this for now, just to actually have it working.

I agree on your validity point as well. I'll get it done.

PR #41, opened 2018-02-13
Edits now use datetime internally.

cmoesgaard:
Refactored terminate implementation to be much less stupid.
Implemented termination of roles, which was missing for whatever reason.
Misc cleanup.

### DIFF
--- a/mora/service/common.py
+++ b/mora/service/common.py
@@ -7,6 +7,7 @@
 #
 import collections
 import copy
+import datetime
 import functools
 from enum import Enum
 from typing import Callable, List, Tuple
@@ -115,11 +116,14 @@ def get_obj_value(obj, path: tuple, filter_fn: Callable = None):
         return props
 
 
-def ensure_bounds(valid_from: str,
-                  valid_to: str,
+def ensure_bounds(valid_from: datetime.datetime,
+                  valid_to: datetime.datetime,
                   props: List[FieldTuple],
                   obj: dict,
                   payload: dict):
+    def filter_fn(prop):
+        return util.from_iso_time(prop['virkning']['from'])
+
     for field in props:
         props = get_obj_value(obj, field.path, field.filter_fn)
         if not props:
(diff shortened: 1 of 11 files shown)


mora/service/common.py
```diff
@@ -115,11 +116,14 @@ def get_obj_value(obj, path: tuple, filter_fn: Callable = None):
...
 
-def ensure_bounds(valid_from: str,
-                  valid_to: str,
+def ensure_bounds(valid_from: datetime.datetime,
+                  valid_to: datetime.datetime,
                   props: List[FieldTuple],
                   obj: dict,
                   payload: dict):
+    def filter_fn(prop):
+        return util.from_iso_time(prop['virkning']['from'])
```
danchr: We should factor this out to a separate function, such as `get_effect_from()` in `common.py` or `util.py`.

mora/service/common.py
```diff
@@ -128,32 +132,32 @@ def ensure_bounds(valid_from: str,
         updated_props = []
         if field.type == FieldTypes.ADAPTED_ZERO_TO_MANY:
             # If adapted zero-to-many, move first and last, and merge
-            sorted_props = sorted(props, key=lambda x: x['virkning']['from'])
+            sorted_props = sorted(props, key=filter_fn)
```
danchr: …and that would also make this code ever so slightly more readable.

mora/service/employee.py
```diff
@@ -288,10 +280,10 @@ def create_employee(employee_uuid):
     """
 
     handlers = {
-        'engagement': create_engagement,
-        'association': create_association,
+        'engagement': engagement.create_engagement,
+        'association': association.create_association,
         'it': itsystem.create_system,
-        'role': create_role,
+        'role': role.create_role,
```
danchr: In general, your reviewer would be ever so slightly happier if you could put trivial changes such as these in a separate commit :)

mora/service/employee.py
```diff
@@ -598,31 +590,22 @@ def terminate_employee(employee_uuid):
...
-        leave_uuid = leave_obj[0]
-        leave.terminate_leave(leave_uuid, date)
+    # Org funks
+    types = [
+        (keys.ENGAGEMENT_KEY, 'Afslut engagement'),
+        (keys.ASSOCIATION_KEY, 'Afslut tilknytning'),
+        (keys.ROLE_KEY, 'Afslut rolle'),
+        (keys.LEAVE_KEY, 'Afslut orlov'),
+    ]
+    for funk_type in types:
```
danchr: You should unwrap this in one line, e.g.:

```python
for key, note in types:
```

Also, I'd suggest using one note for all actions — after all, they are infact separate results of the same user intent, and perhaps tuple since you don't need mutability.

mora/service/employee.py
cmoesgaard: Changes like these usually happen as a part of fixing other things and aren't necessarily easy to always separate from the rest, but I agree with your point.
I'll try to split my commits into logical chunks in the future. :)

mora/service/common.py
```diff
@@ -128,32 +134,32 @@ def ensure_bounds(valid_from: str,
...
             # If adapted zero-to-many, move first and last, and merge
-            sorted_props = sorted(props, key=lambda x: x['virkning']['from'])
+            sorted_props = sorted(props, key=get_effect_from)
             first = sorted_props[0]
             last = sorted_props[-1]
 
             # Check bounds on first
-            if valid_from < first['virkning']['from']:
-                first['virkning']['from'] = valid_from
+            if valid_from < util.parsedatetime(first['virkning']['from']):
```
danchr: → `get_effect_from()`

mora/service/common.py
```diff
@@ -128,32 +134,32 @@ def ensure_bounds(valid_from: str,
...
+                first['virkning']['from'] = util.to_lora_time(valid_from)
                 updated_props = sorted_props
-            if last['virkning']['to'] < valid_to:
-                last['virkning']['to'] = valid_to
+            if util.parsedatetime(last['virkning']['to']) < valid_to:
```
danchr: → `get_effect_to()`

PR #43, opened 2018-02-13
Feature/20348 leder

cmoesgaard:
Implemented reading and writing of Manager in the usual fashion.
The code for creating new org funk objects has become even more contrived due to having to special handling of 'opgaver' -- it works for now, but will be overhauled in the near future hopefully, along with all of the nearly identical code in the edit functions for each org funk type.

### DIFF
--- a/mora/service/common.py
+++ b/mora/service/common.py
@@ -89,13 +89,9 @@ def inactivate_old_interval(old_from: str, old_to: str, new_from: str,
     return payload
 
 
-def set_object_value(obj: dict, path: tuple, val: Union[dict, List[dict]],
-                     overwrite: bool = False):
+def set_object_value(obj: dict, path: tuple, val: List[dict]):
     path_list = list(path)
     obj_copy = copy.deepcopy(obj)
-
-    if val is not list:
-        val = list(val)
 
     current_value = obj_copy
     while path_list:
@@ -103,7 +99,7 @@ def set_object_value(obj: dict, path: tuple, val: Union[dict, List[dict]],
         if path_list:
             current_value = current_value.setdefault(key, {})
         else:
-            if overwrite or not current_value.get(key):
+            if not current_value.get(key):
                 current_value[key] = val
             else:
                 current_value[key].extend(val)
--- a/mora/service/common.py
+++ b/mora/service/common.py
@@ -10,7 +10,7 @@ import copy
 import datetime
 import functools
 from enum import Enum
-from typing import Callable, List, Tuple
+from typing import Callable, List, Tuple, Union
 
 import flask
 import iso8601
@@ -89,10 +89,13 @@ def inactivate_old_interval(old_from: str, old_to: str, new_from: str,
     return payload
 
 
-def set_object_value(obj: dict, path: tuple, vals: List[dict],
+def set_object_value(obj: dict, path: tuple, val: Union[dict, List[dict]],
                      overwrite: bool = False):
     path_list = list(path)
     obj_copy = copy.deepcopy(obj)
+
+    if val is not list:
+        val = list(val)
 
     current_value = obj_copy
     while path_list:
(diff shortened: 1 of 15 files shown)


mora/service/common.py
```diff
@@ -89,21 +89,24 @@ def inactivate_old_interval(old_from: str, old_to: str, new_from: str,
     return payload
 
 
-def set_object_value(obj: dict, path: tuple, vals: List[dict],
+def set_object_value(obj: dict, path: tuple, val: Union[dict, List[dict]],
                      overwrite: bool = False):
     path_list = list(path)
     obj_copy = copy.deepcopy(obj)
 
+    if val is not list:
```
danchr: This can't possibly work — don't you mean:

```python
if not isinstance(val, list):
```

…but that suggests you're probably not hitting this case in the tests, so you should either adapt them or drop this clause.

mora/service/engagement.py
```diff
@@ -43,14 +43,15 @@ def list_details(type, id):
...
-    The value above informs you that 'association', 'engagement', 'role' and
-    'leave' are valid for this entry, and that no entry exists at any time
-    for 'association' and 'role', whereas 'engagement' and 'leave' have at
-    least one entry either in the past, present or future.
-
+    The value above informs you that 'association', 'engagement', 'role',
+    'leave' and 'manager' are valid for this entry, and that no entry exists
+    at any time for 'association', 'role' and 'manager', whereas
+    'engagement' and 'leave' have at least one entry either in the past,
+    present or future.
```
danchr: This is getting rather long-winded, I'd say 🙂 Perhaps we should rephrase it to just describe one or two `true` values, and describe everything else as “the others”…

mora/service/engagement.py
```diff
@@ -266,28 +281,43 @@ def convert_leave(funcid, effect):
...
             keys.FROM: util.to_iso_time(start),
             keys.TO: util.to_iso_time(end),
         }
         return func
 
+    def get_classes(effect):
+        rels = effect['relationer']
+        return [obj.get('uuid') for obj in itertools.chain(
+            rels.get('opgaver', []),
+            rels.get('organisatoriskfunktionstype', [])
```
danchr: This is in fact fine, but that's not obvious from reading the code. One might think that e.g. `RESPONSIBILITY_FIELD` refers to something other than, say, `MANAGER_LEVEL_FIELD`, but they do not. All these functions are just begging for a `dict` that configures this…

mora/service/org.py
```diff
@@ -142,7 +147,16 @@ def get_organisation(orgid):
...
+    engagements = c.organisationfunktion(tilknyttedeorganisationer=orgid,
+                                         funktionsnavn=keys.ENGAGEMENT_KEY)
+    associations = c.organisationfunktion(tilknyttedeorganisationer=orgid,
+                                          funktionsnavn=keys.ASSOCIATION_KEY)
+    leaves = c.organisationfunktion(tilknyttedeorganisationer=orgid,
+                                    funktionsnavn=keys.LEAVE_KEY)
+    roles = c.organisationfunktion(tilknyttedeorganisationer=orgid,
+                                   funktionsnavn=keys.ROLE_KEY)
+    managers = c.organisationfunktion(tilknyttedeorganisationer=orgid,
+                                      funktionsnavn=keys.MANAGER_KEY)
```
danchr: We'll have to keep our eye out for how this affects performance. In particular, do the SQL stored procedures perform the same array-compare when filtering for `funktionsnavn` as they do when filtering for `gyldighed`? If so, we can't afford this granularity, as the cost is prohibitive when counting the 36k engagements in the Aarhus spreadsheet.

cmoesgaard: That's a good point. I'll let it stay for now, but keep in mind for next time we run it with a large dataset.
But let's also discuss later whether or not this level of granularity is actually useful information at all

danchr: Yeah, the idea is just to present *something* — and let someone else decide whether it's useful at all…

PR #44, opened 2018-02-14
Address reading support

danchr:
This change series adds support for reading addresses, and fixes various other minor issues I stumbled into.

Posting this now before it reaches an even more unmanageable size…


mora/converters/reading.py
```diff
@@ -267,8 +267,8 @@ def convert_address(obj):
                 'location': _get_location(info.location),
                 'visibility': {
                     'user-key': info.visibility,
-                    'uuid': meta.PHONE_VISIBILITY_UUIDS[info.visibility],
-                    'name': meta.PHONE_VISIBILITIES[info.visibility],
+                    'uuid': meta.PHONE_VISIBILITY_UUIDS.get(info.visibility),
+                    'name': meta.PHONE_VISIBILITIES.get(info.visibility),
```
cmoesgaard: Are we still using things from this file for the new API? If so, we should extract the relevant methods so we can easily this kill this entire file along with the rest of the legacy API when the time is right.

mora/service/common.py
```diff
@@ -26,6 +27,16 @@ class FieldTypes(Enum):
     ADAPTED_ZERO_TO_MANY = 2,
 
 
+class AbstractRelationDetail(abc.ABC):
```
cmoesgaard: Overall, I like this idea.
I'm guessing the end goal is having separate strategies for each type of detail, i.e. engagement, association etc.?

mora/service/engagement.py
```diff
@@ -94,10 +94,15 @@ def get_systems(reg):
 )
 @util.restrictargs('at', 'validity', 'start', 'limit')
 def get_engagement(type, id, function):
```
cmoesgaard: One of these days we should really rename this function and move it to somewhere more appropriate.

mora/service/common.py
danchr: Generally speaking, the point is that IT systems and addresses are similar, yet different from the others. Eventually, yes, it might make sense to have each service as a simple class.

mora/converters/reading.py
danchr: No, but the changes to the fixtures causes us to have ‘visibilities’ which aren't, in fact, that. So it was just extra error handling for now. Probably should've mentioned that in the commit…

mora/service/engagement.py
danchr: Indeed, yes; the actual design turned out somewhat different from I had envisioned ;)

mora/service/common.py
cmoesgaard: Okay, good. I might be trying to plan a bit too far ahead for when we potentially have to support a number of slightly different implementations for each municipality -- but I'd really like, sooner rather than later, to extract all of the specific logic, in an open/closed fashion to make it as easy as possible to swap out all the specific implementation details. This is a good step in that direction at least.

danchr: Yeah, we're in agreement here — we want to keep moving slowly but steadily in that direction 🙂

PR #45, opened 2018-02-15
Feature/org unit writing

cmoesgaard:
Implemented the usual writing magic for org units.
Addresses aren't handled as of yet - awaiting final implementation of address writing logic.
Renamed some of the mapping fields to something slightly more specific

### DIFF
--- a/mora/service/org.py
+++ b/mora/service/org.py
@@ -17,11 +17,16 @@ organisational units.
 
 import enum
 import operator
+import uuid
 
 import flask
 import werkzeug
 
-from . import common, keys
+from mora import lora
+from . import mapping
+from .common import (create_organisationsenhed_payload, _create_virkning,
+                     inactivate_old_interval, set_object_value, update_payload,
+                     ensure_bounds)
 from .. import util
 from . import common
 from . import keys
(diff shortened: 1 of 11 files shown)


tests/util.py
```diff
@@ -357,6 +359,33 @@ def _perform_request(self, path, **kwargs):
...
+        list ordering
+        """
+        obj = copy.deepcopy(obj)
+        for outer_value in filter(lambda x: type(x) is dict, obj.values()):
+            for inner_key in outer_value:
+                outer_value[inner_key] = sorted(outer_value[inner_key],
+                                                key=common.get_effect_from)
+        return obj
+
+    def assertEqualLoRa(self, expected, actual):
```
danchr: “LoRA” isn't terribly helpful in this regard — perhaps `assertRegistrationsEqual` instead? (That appears to be what you're actually want to compare.) Also, a docstring would be nice.

tests/util.py
```diff
@@ -357,6 +359,33 @@ def _perform_request(self, path, **kwargs):
 
         return self.client.open(path, **kwargs)
 
+    def sort_inner_lists(self, obj):
```
danchr: This method doesn't actually refer to or use `self` anywhere — I'd say this should be a regular function instead. If you only use it within the other method, just having it as an internal helper function should suffice.

tests/util.py
```diff
@@ -357,6 +359,33 @@ def _perform_request(self, path, **kwargs):
...
+
+        This is purely to help comparison tests, as we don't care about the
+        list ordering
+        """
+        obj = copy.deepcopy(obj)
+        for outer_value in filter(lambda x: type(x) is dict, obj.values()):
+            for inner_key in outer_value:
+                outer_value[inner_key] = sorted(outer_value[inner_key],
+                                                key=common.get_effect_from)
+        return obj
```
danchr: Personally, I'd use a `dict`-comprehension here, but that's just me 🙂

mora/service/org.py
```diff
@@ -17,11 +17,16 @@
...
 
 import flask
 import werkzeug
 
-from . import common, keys
+from mora import lora
+from . import mapping
+from .common import (create_organisationsenhed_payload, _create_virkning,
+                     inactivate_old_interval, set_object_value, update_payload,
+                     ensure_bounds)
```
danchr: I don't know why, but this style of import always seemed a bit messy to me…

tests/util.py
cmoesgaard: Yeah, I'm terrible at naming things. I'll fix it.

mora/service/org.py
cmoesgaard: I agree, that's why I fixed it. :)

danchr: Yeah, I noticed that later 🙂

PR #47, opened 2018-02-16
Added data for manager to BALLERUP

cmoesgaard:
Updated import process to handle lederniveau and lederansvar


tests/test_integration_importing.py
```diff
@@ -1203,6 +1218,130 @@ def get(f):
                          'user_key': 'Barselsorlov',
                          'uuid': 'd7ec3a18-3a9d-43c8-ad03-0202c0d044d4'}
                     ],
+                    'manager_level': [
+                        {'example': None,
+                         'name': 'Niveau 90',
```
danchr: (90 is not a sufficiently random number; numbers such as 7, 42 and 1337 are vastly superior in that regard.)

PR #48, opened 2018-02-19
Address writing support (#21288)

danchr:
This is a rather large series; let me know if it's too much and I'll break it up.


mora/service/address.py
```diff
@@ -147,9 +149,110 @@ def convert(start, end, effect):
...
+        'relationer': {
+            'adresser': orig_addresses + new_addreses,
+        },
+        'note': 'Tilføj adresse',
+    }
+
+    c.bruger.update(payload, employee_uuid)
+
+
+def edit_address(employee_uuid, req):
```
cmoesgaard: Ditto my comment for create_address

mora/service/address.py
```diff
@@ -147,9 +149,110 @@ def convert(start, end, effect):
...
+            'urn': addr.URN_FORMATS[scope].format(value),
+            'objekttype': classobj['uuid'],
+            'virkning': effect,
+        }
+
+    else:
+        raise ValueError('unknown address scope {!r}!'.format(scope))
+
+
+def create_address(employee_uuid, req):
```
cmoesgaard: We should probably decide on whether to do it like this, or like creating IT systems, and generalize it, considering how similar the procedures are.

danchr: Agreed, I intend to do so in a followup change.

PR #50, opened 2018-02-21
Feature/history

cmoesgaard:
Implemented history for org units and employees.
History for employees, as discussed, has been implemented using some iffy pseudo no-op writes to the users whenever a new object is associated to them, or one of their associated objects changes.
I've denoted the various places in the code affected by this less-than-ideal approach, so we can hopefully go back and fix this in the future when LoRa allows it.

The information returned from the history endpoints is as close to what we recieved from the old API as I felt made sense.


mora/service/common.py
```diff
@@ -506,3 +516,36 @@ def is_reg_valid(reg):
     return any([gyldighed_obj.get('gyldighed') == 'Aktiv'
                 for tilstand in reg.get('tilstande', {}).values()
                 for gyldighed_obj in tilstand])
+
+
+def add_bruger_history_entry(employee_uuid, note: str):
```
danchr: This could use a docstring, perhaps including the comment below.

mora/service/common.py
```diff
@@ -17,15 +17,25 @@
...
+
+RELATION_TRANSLATIONS = {
+    'engagement': keys.ENGAGEMENT_KEY.lower(),
+    'association': keys.ASSOCIATION_KEY.lower(),
+    'it': keys.ITSYSTEM_KEY.lower(),
+    'role': keys.ROLE_KEY.lower(),
+    'address': keys.ADDRESS_KEY.lower(),
+    'manager': keys.MANAGER_KEY.lower(),
+    'leave': keys.LEAVE_KEY.lower(),
+}
```
danchr: This feels a bit too specific — don't we have an existing map that almost does the same?

mora/service/common.py
```diff
@@ -506,3 +516,36 @@ def is_reg_valid(reg):
...
+        'note': note
+    }
+
+    payload = set_object_value(payload, path, [gyldighed])
+    c.bruger.update(payload, employee_uuid)
+
+
+def convert_reg_to_history(reg):
+    return {
+        'user_ref': reg['brugerref'],
```
danchr: Eventually, we'll want to look up this user object. LoRA has a particular UUID which means ‘anonymous’ — 😒 — which we should preferably detect.

mora/service/employee.py
```diff
@@ -413,6 +415,12 @@ def create_employee(employee_uuid):
 
         handler(str(employee_uuid), req)
 
+        # Write a noop entry to the user, to be used for the history
+        common.add_bruger_history_entry(
+            employee_uuid,
+            "Opret {}".format(common.RELATION_TRANSLATIONS[role_type])
```
danchr: I think we should consider making the notes even more verbose and descriptive — but not necessarily right now.

mora/service/employee.py
```diff
@@ -806,7 +822,7 @@ def terminate_employee(employee_uuid):
     """
     date = common.get_valid_from(flask.request.get_json())
 
-    c = lora.Connector(effective_date=date)
+    c = lora.Connector()
```
danchr: Is this safe?

mora/service/keys.py
```diff
@@ -20,6 +20,7 @@
 JOB_FUNCTION = 'job_function'
 ITSYSTEM = 'itsystem'
 ADDRESS = 'address'
+ITSYSTEM_KEY = 'IT system'
```
danchr: “IT-system”

tests/test_integration_history.py
```diff
@@ -0,0 +1,308 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+
+from flask import json
+
+from . import util
+
+
+class HistoryTest(util.LoRATestCase):
```
danchr: Our previous history tests ran each test as its own class — in order to ensure that they started out with a clean slate. Do these tests still pass if you run them in reverse order?

mora/service/common.py
cmoesgaard: Yeah, I agree - this is just more or less copying the approach used for history entries in the old API.

mora/service/employee.py
cmoesgaard: Ideally we would add specific information regarding what was changed, but that probably requires rewriting a bunch of logic, as that information is not readily available.

mora/service/employee.py
cmoesgaard: Nope. Should probably have a test to ensure this.

tests/test_integration_history.py
cmoesgaard: They operate on different objects, so they're safe.
Runnings each of them in separate classes is still a good idea, though. I'll fix that.

danchr: As an optimisation, we don't actually create a full LoRA instance for test, but rather for each class. Then, we simply iterate over all relevant objects between runs and `DELETE` them. This is fine, as long as we don't happen to look at the registration history of each object — which is true for everything other than our history support. Hence my question…

mora/service/common.py
cmoesgaard: We have a map containing all of the keys for the org funk objects, yes. Extending that with 'address' and 'it' breaks stuff, though.

tests/util.py
```diff
@@ -378,9 +378,13 @@ def sort_inner_lists(obj):
...
-        del actual['fratidspunkt'], actual[
-            'tiltidspunkt'], actual[
-            'brugerref']
+        expected.pop('fratidspunkt', None)
+        expected.pop('tiltidspunkt', None)
+        expected.pop('brugerref', None)
+
+        actual.pop('fratidspunkt', None)
+        actual.pop('tiltidspunkt', None)
+        actual.pop('brugerref', None)
```
danchr: Cosmetic, i know, but you could simplify this with:

```python
expected.update(fratidspunkt=None, tiltidspunkt=None, brugerref=None)
actual.update(fratidspunkt=None, tiltidspunkt=None, brugerref=None)
```

Doesn't `brugerref` remain the same?

PR #51, opened 2018-02-21
Time machine

cmoesgaard:
Added some tests and a fix for an edge case where we attempt to get an object outside of the validity period of its properties, but where metadata still exists at that given time.

### DIFF
--- a/mora/service/common.py
+++ b/mora/service/common.py
@@ -493,3 +493,15 @@ def replace_relation_value(relations: List[dict],
 
     else:
         raise ValueError('original entry not found!')
+
+
+def is_reg_valid(reg, gyldighed_key):
+    """
+    Check if a given registration is valid
+
+    :param reg: A registration object
+    :param gyldighed_key: A key for where to find the gyldighed on the object
+    :return: True if registration 'gyldighed' is 'Aktiv'
+    """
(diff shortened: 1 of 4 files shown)


mora/service/common.py
```diff
@@ -459,3 +459,15 @@ def get_valid_to(obj, fallback=None):
...
+def is_reg_valid(reg, gyldighed_key):
+    """
+    Check if a given registration is valid
+
+    :param reg: A registration object
+    :param gyldighed_key: A key for where to find the gyldighed on the object
+    :return: True if registration 'gyldighed' is 'Aktiv'
+    """
+    return bool(get_obj_value(reg, ('tilstande', gyldighed_key),
+                              lambda x: x.get('gyldighed') == 'Aktiv'))
```
danchr: Couldn't you simplify this code a bit by iterating over the states, and check if any are active?

cmoesgaard: That is essentially what's happening already. But it could probably be simplified a bit.

danchr: I mean that you could iterate over all states rather than specify `gyldighed_key` — that seems a bit cleaner to me?

cmoesgaard: That's a good point

PR #54, opened 2018-02-26
Merge development into master

danchr:


danchr:
> I suggest we try to follow git flow to some extent at least, and create a release branch, instead of merging this directly to master.

> That allows us to test on a prospective release and fix any eventual bugs we might encounter.

How eminently reasonable. We just need to figure out how to integrate that with our test server…

PR #55, opened 2018-02-27
Moved and refactored old address lookup functionality

cmoesgaard:


mora/service/address.py
```diff
@@ -256,3 +262,78 @@ def edit_address(employee_uuid, req):
     }
 
     c.bruger.update(payload, employee_uuid)
+
+
+@blueprint.route('/address_lookup/')
```
danchr: I think we should move this to `/o/<orgid>/` — both for consistency and clarity.

mora/service/address.py
```diff
@@ -256,3 +262,78 @@ def edit_address(employee_uuid, req):
...
+@util.restrictargs('local', required=['query'])
```
danchr: Regardless, this definitely shouldn't be called `local` — `org` or something similar is better. I also think we should a simple and customary `q` for the query parameter.

mora/service/address.py
```diff
@@ -256,3 +262,78 @@ def edit_address(employee_uuid, req):
...
+def get_geographical_addresses():
```
danchr: Rename to something containing ‘autocomplete’.

mora/service/address.py
```diff
@@ -256,3 +262,78 @@ def edit_address(employee_uuid, req):
...
+    **Example Response**:
+
+    :<jsonarr uuid uuid: A UUID of a DAR address
+
+    .. sourcecode:: json
+
+      [
+        {
+          "location": {
+            "uuid": "397705eb-1752-4efc-8f1c-54fd0d76366c"
```
danchr: Having a UUID is quite nice, but we shouldn't omit the ever so slightly more readable actual address representation.

mora/service/address.py
```diff
@@ -256,3 +262,78 @@ def edit_address(employee_uuid, req):
...
+          }
+        },
+        {
+          "location": {
+            "uuid": "37380cfe-3781-460e-95f0-36ad8798f5ea"
+          }
+        }
+      ]
+    """
+    return flask.jsonify(
```
danchr: This seems a bit too much code to me — couldn't you just inline the functions below?

mora/service/address.py
cmoesgaard: Having the endpoint be located at `/o/<orgid>/` might make the user believe that the address search is automatically local to that specific organisation, which it isn't.

danchr: The `local` argument actually works that way — mostly, anyway — and the UI actually does work that way.

cmoesgaard: My point is, it would then be possible to search for an address on `/o/<org 1>/address_lookup?q=asd&orgid=<org 2>`, which would search for addresses in org 2, not org 1.

danchr: Sorry, perhaps I wasn't clear; what I suggested was:

    /o/<org 1>/address_lookup?q=asd

_Instead_ of the `orgid`/`local` parameter. Then, in order to search for adresses in the entire country, you'd do:

    /o/<org 1>/address_lookup?q=asd&global=1

Or something similar…

cmoesgaard: Right, that makes sense. I'll change it.

danchr: Great 🙂 I think also forgot to mention this earlier, but I'd suggest you include ‘autocomplete’ in the URL endpoint — it's what we use it for.

mora/service/org.py
```diff
@@ -761,3 +765,77 @@ def get_org_unit_history(unitid):
...
+        {
+          "location": {
+            "uuid": "0a3f50cb-05eb-32b8-e044-0003ba298018"
+            "name": "Wild Westvej 1, 9310 Vodskov"
+          }
+        }
+      ]
+    """
+    q = flask.request.args['q']
+    global_lookup = flask.request.args.get('orgid')
```
danchr: `s/orgid/global/`? I think we need a test for this 🙂 Also, we might want to do a bit more than merely check for presence and truth value. For example:

    is_global = flask.request.args.get('global', False, bool)

Not sure which works best, `bool` or `int`, but Werkzeug's docs has [this](http://werkzeug.pocoo.org/docs/0.14/datastructures/#werkzeug.datastructures.TypeConversionDict.get) and [this](http://werkzeug.pocoo.org/docs/0.14/datastructures/#werkzeug.datastructures.MultiDict.get) to offer.

tests/test_integration_service.py
```diff
@@ -1115,3 +1116,138 @@ def test_detail_list(self):
...
+            '/service/o/00000000-0000-0000-0000-000000000000/'
+            'address_autocomplete/?q=42',
+            {
+                'message': 'No local municipality found!',
+                'status': 404,
+            },
+            status_code=404,
+        )
+
+    @patch('mora.service.org.requests')
```
danchr: `s/mora.service.org//`? Specifying the value after import seems odd. The original module should work quite well in most cases, unless someone's doing something silly — in which case this probably won't work either…

cmoesgaard: I'm not sure what you mean by specifying the value after import.

My idea here is to be as specific as possible in my mocking, as apposed to just blanket-mocking all the uses of `requests`, which could lead to potential problems in our use of `requests` in `lora.py`

danchr: Oh, I see! For what it's worth I added an `allow_mox` argument to `mock()` in 320b905e2bd768a7e3969f6c4470810bf3180a79 so that you can mock everything other than LoRA. It does the same, but based on URL prefixes instead.

cmoesgaard: I've updated the tests to use `mock()`. The `allow_mox` argument doesn't play well with minimox, it seems, though - but fixing that seemed a bit out of scope for this.

mora/service/org.py
```diff
@@ -761,3 +765,77 @@ def get_org_unit_history(unitid):
...
+            "uuid": "0a3f50cb-05eb-32b8-e044-0003ba298018"
+            "name": "Wild Westvej 1, 9310 Vodskov"
+          }
+        }
+      ]
+    """
+    q = flask.request.args['q']
+    global_lookup = flask.request.args.get('global', False, type=bool)
+
+    if not global_lookup:
```
danchr: AFAICT the tests don't cover this case.

mora/service/org.py
```diff
@@ -761,3 +765,77 @@ def get_org_unit_history(unitid):
                                unit_registrations))
 
     return flask.jsonify(history_entries)
+
+
+@blueprint.route('/o/<uuid:orgid>/address_autocomplete/')
```
danchr: Minor nit: I was a bit surprised to see this in the `org` module — doesn't it belong to `address`?

(Also, having it here triggers a merge conflict due to my recent changes.)

tests/test_integration_service.py
```diff
@@ -1115,3 +1116,211 @@ def test_detail_list(self):
                     'role': False,
                 },
             )
+
+
+class TestAddressLookup(util.LoRATestCase):
```
danchr: Minor nit: this test no longer accesses LoRA directly, so it isn't an integration test, anymore, right? As such, you can make the test suite run ever so slightly faster by descending from `util.TestCase` instead, and preferably also moving it to another file — for consistency.

mora/service/org.py
cmoesgaard: Both global and non-global lookup is handled by tests.

mora/service/org.py
cmoesgaard: I moved it to `org` because of the change in endpoint address - I'm not sure how that made sense to me. I'll change it back.

tests/test_integration_service.py
```diff
@@ -1115,3 +1116,211 @@ def test_detail_list(self):
                     'role': False,
                 },
             )
+
+
+class TestAddressLookup(util.TestCase):
```
danchr: This should be in a separate file, such as `test_service_address.py`.

PR #56, opened 2018-02-27
Fix bug #21726, among others

danchr:
This fixes bug [#21726](https://redmine.magenta-aps.dk/issues/21726) — what I wrote for org. units couldn't possibly work.


mora/service/orgunit.py
```diff
@@ -0,0 +1,647 @@
...
+
+    :status 404: If the organisational unit isn't found.
+    :status 200: Otherwise.
+
+    **Example Response**:
+
+    .. sourcecode:: json
+
+      [
+        {
```
cmoesgaard: We should also return `org_unit_type` and `parent`

danchr: Good point! I'll add that in a followup so that @n-singh has something to work with :)

tests/test_integration_service.py
```diff
@@ -332,15 +332,15 @@ def test_orgunit(self):
         self.load_sample_structures(minimal=True)
 
         with self.subTest('invalid'):
-            self.assertRequestFails(
+            self.assertRequestResponse(
                 '/service/ou/00000000-0000-0000-0000-000000000000/',
-                404,
+                [],
```
cmoesgaard: If we return an empty list on an invalid UUID, the docstring for the endpoint should be updated accordingly.

PR #58, opened 2018-02-28
orgunit: let /details/info redirect to the unit itself

danchr:
A minimal change to fulfil Novik's request.


mora/service/engagement.py
```diff
@@ -98,7 +98,7 @@ def list_details(type, id):
     '/<any("e", "ou"):type>/<uuid:id>/details/<function>',
 )
 @util.restrictargs('at', 'validity', 'start', 'limit')
-def get_engagement(type, id, function):
+def get_detail(type, id, function):
```
cmoesgaard: I think it's actually time to move this to a file that isn't `engagement.py`

danchr: Really? What a novel suggestion 😏

PR #59, opened 2018-02-28
#21726 — fix orgunit

danchr:
This change adjusts the org. unit API to better correspond to the wishes of the frontenders.


mora/service/orgunit.py
```diff
@@ -63,23 +100,26 @@ def get_one_orgunit(c, unitid, unit=None,
...
     elif details is UnitDetails.NCHILDREN:
+        r.update(base_info)
+
         children = c.organisationenhed(overordnet=unitid, gyldighed='Aktiv')
 
         r['child_count'] = len(children)
 
-    else:
+    elif details is UnitDetails.FULL:
+        r[keys.ORG_UNIT] = base_info
```
cmoesgaard: It needs to be clear that these are properties of the object itself, and not relations to other objects.

cmoesgaard: Having the objects properties nested one level down is also inconsistent with how every other detail object is structured.

danchr: Really? It's an explicit request from Anders…

danchr: (And I agree about your first point — I just wasn't sure whether it affected other objects as well.)

danchr: Agreed.

PR #63, opened 2018-03-05
#21288 addresses — both frontend & backend

danchr:
This pull request fixes various issues I ran into last Friday. With this, addresses and organisational units work as expected, and render correctly in the UI.

As an aside, I added support for linking between objects — and addresses too.


tests/test_integration_service.py
```diff
@@ -1084,6 +1039,7 @@ def test_detail_list(self):
                     'it': True,
                     'leave': False,
                     'manager': False,
+                    'org_unit': False,
```
cmoesgaard: This is a bit silly. 

Following my previous comment, I don't think we should have the same set of relation details available for both org units and employees.

danchr: Eh, which comment would that be?

cmoesgaard: A comment on commit `0a4cbac`. It shows up as 'outdated' because of your merge, I guess.

mora/service/employee.py
```diff
@@ -36,6 +37,13 @@
                             url_prefix='/service')
 
 
+RELATION_TYPES = {
+    'it': itsystem.ITSystems,
+    'address': address.Addresses,
+    'org_unit': orgunit.OrgUnit,
```
cmoesgaard: This allows us to now potentially create and edit org units through the employee endpoints. Does that make sense?

PR #65, opened 2018-03-07
Implemented new response from facet endpoints

cmoesgaard:


mora/service/facet.py
```diff
@@ -230,13 +245,20 @@ def get_classes(orgid: uuid.UUID, facet: str):
     limit = int(flask.request.args.get('limit') or 1000)
```
danchr: While at it, perhaps you should update this to use our new paging support?

cmoesgaard: Good idea.

Something along these lines?
```
{
    "name": "address_type",
    "path": "/service/o/456362c4-0ee4-4e5e-a72c-751239745e62/f/address_type/",
    "user_key": "Adressetype",
    "uuid": "e337bab4-635f-49ce-aa31-b44047a43aa1",
    "data": {
        "items": [
            {
                "example": "http://www.korsbaek.dk/",
                "name": "Hjemmeside",
                "scope": "WWW",
                "user_key": "URL",
                "uuid": "160ecaed-50b0-4800-bebc-0d0289a4f624"
            },
            {
                "example": "<UUID>",
                "name": "Lokation",
                "scope": "DAR",
                "user_key": "AdresseLokation",
                "uuid": "031f93c3-6bab-462e-a998-87cad6db3128"
            },
            {
                "example": "Mandag 10:00-12:00 Tirsdag 14:00-16:00",
                "name": "Åbningstid, telefon",
                "scope": "TEXT",
                "user_key": "Åbningstid Telefon",
                "uuid": "0836ffbf-3b3e-410f-8cbf-face7e6844ef"
            }
        ],
        "offset": 0,
        "total": 3
    }
}
```
Or should we move the pagination metadata out to the same level as everything else?

danchr: I can see arguments either way…

danchr: On second though, I think what you posted is best — both when it comes to implementation and ease of understanding.

danchr:
FWIW your JSON isn't valid, and you have two extraneous imports:

http://lorajenkins.atlas.magenta.dk/job/mora/job/feature%252F21391_facetter/3/warnings58Result/
http://lorajenkins.atlas.magenta.dk/job/mora/job/feature%252F21391_facetter/3/warnings46Result/

PR #68, opened 2018-03-08
#21708 address support for managers

danchr:


cmoesgaard:
What about creating addresses for managers?

danchr:
> What about creating addresses for managers?

I'm not sure I follow — what do you mean?

cmoesgaard:
From what I can see, you've implemented editing addresses for managers, but it's not possible to actually create addresses, as part of the process of creating a manager.

danchr:
@cmoesgaard Ping?

PR #70, opened 2018-03-12
We now only get org units that have a gyldighed of 'Aktiv', behavling…

cmoesgaard:
… similarly to fetching other details.

Updated terminate test to verify this, and to also more thoroughly test functionality using the get details endpoint instead of hitting LoRa directly.

### DIFF
--- a/mora/service/orgunit.py
+++ b/mora/service/orgunit.py
@@ -81,7 +81,10 @@ class OrgUnit(common.AbstractRelationDetail):
                     ),
                 },
             )
-            if c.is_effect_relevant({'from': start, 'to': end})
+            if c.is_effect_relevant({'from': start, 'to': end}) and
+            effect.get('tilstande')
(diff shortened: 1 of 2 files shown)


mora/service/orgunit.py
```diff
@@ -81,7 +81,10 @@ def get(self, objid):
                     ),
                 },
             )
-            if c.is_effect_relevant({'from': start, 'to': end})
+            if c.is_effect_relevant({'from': start, 'to': end}) and
+            effect.get('tilstande')
+                  .get('organisationenhedgyldighed')[0]
+                  .get('gyldighed') == 'Aktiv'
```
danchr: I think we have loads of statements like this, and it's actually kind of horrible. Would be nice to do `c.has_state(effect, 'Aktiv')` instead…

cmoesgaard: Yes, I agree. Let's do that at some point...

PR #71, opened 2018-03-13
#21879 orgunit addresses

danchr:
This implements the server-side logic for specifying addresses when creating of organisational units. While at it, I cleaned up the handling of phone numbers ever so slightly.

### DIFF
--- a/mora/service/address.py
+++ b/mora/service/address.py
@@ -18,21 +18,31 @@ import itertools
 import re
 
 import flask
-import parse
 import requests
 
 from .. import lora
 from .. import util
-from ..converters import addr
 
 from . import common
 from . import facet
 from . import keys
 
-HREF_FORMATS = {
-    'EMAIL': 'mailto:{}',
-    'PHONE': 'tel:+45{}',
-    'WWW': '{}',
+session = requests.Session()
+session.headers = {
+    'User-Agent': 'MORA/0.1',
+}
+
+URN_PREFIXES = {
+    'EMAIL': 'urn:mailto:',
+    'PHONE': 'urn:magenta.dk:telefon:',
+    'EAN': 'urn:magenta.dk:ean:',
+    'WWW': 'urn:magenta.dk:www:',
+}
+
+HREF_PREFIXES = {
+    'EMAIL': 'mailto:',
+    'PHONE': 'tel:',
+    'WWW': '',
 }
 
 MUNICIPALITY_CODE_PATTERN = re.compile('urn:dk:kommune:(\d+)')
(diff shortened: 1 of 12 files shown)


mora/service/address.py
```diff
@@ -82,9 +106,21 @@ def get_address_class(c, addrrel, class_cache):
 
 def get_one_address(c, addrrel, class_cache=None):
     addrclass = get_address_class(c, addrrel, class_cache)
+    addrformat = addrclass['scope']
```
cmoesgaard: ...And the same goes for this function as well, basically.

tests/test_integration_address.py
```diff
@@ -675,24 +674,209 @@ def test_create_unit_address(self):
         # NOW CREATE IT
 
         self.assertRequestResponse(
-            '/service/ou/{}/create'.format(unitid),
+            '/service/ou/{}/edit'.format(unitid),
```
cmoesgaard: Is edit used to create addresses for org units?

mora/service/address.py
```diff
@@ -55,9 +68,20 @@ def get_relation_for(classobj, value):
...
+
+        if scope == 'PHONE':
+            value = re.sub(r'\s+', '', value)
+
+            if not value.startswith('+'):
+                value = '+45' + value
+
+        if not util.is_urn(value):
+            value = prefix + value
+
```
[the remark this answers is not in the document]
danchr: Well, the thing is that it'll be hard to find a format language that's sufficiently expressive; for example, we'll want to format some values as numbers, with zero padding, or prepend a country code to a phone number unless already specified.

Hence, I think certain hardcoded formats combined with a list of URN prefixes is better.

tests/test_integration_address.py
danchr: From my understanding, `create` is used to create the unit itself, whereas all details are modified using the `edit` endpoint. Still, this test doesn't actually manage to add any addresses — considering that editing doesn't work yet — so I'll remove it.

cmoesgaard: As far as I can see, we use the `create` endpoint for creating addresses for employees, so I think we should try to along the workflow for `e` and `ou` as much as we can.

mora/service/address.py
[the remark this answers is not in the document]
danchr: Alas, we have to have _two_ functions: one to append the prefix, and another one to remove it. Plus a third one for generating the URL. On the other hand, we really do have the case that most of our URNs are quite simple — apart from phone numbers.

I'd like to fix it, I just haven't had a good idea, yet.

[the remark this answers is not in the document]
cmoesgaard: I'm at least thinking we'll reach a point where we have to implement specific address types for different customers.

I'm not necessarily saying it's something we should fix now - but I think we should at least start planning ahead for this do be done _at some point_.

PR #72, opened 2018-03-15
#21879 editing & creation of addresses for organisational units

danchr:


mora/service/orgunit.py
```diff
@@ -424,119 +506,157 @@ def edit_org_unit(unitid):
...
+          }
+        }
+      ]
+
+    """
+
+    reqs = flask.request.get_json()
+
+    if not isinstance(reqs, list):
+        return flask.jsonify('Root object must be a list!'), 400
```
cmoesgaard: Why not do the same thing here, of wrapping the payload in a list, if it isn't?

mora/service/orgunit.py
```diff
@@ -424,119 +506,157 @@ def edit_org_unit(unitid):
 
     :statuscode 200: The edit succeeded.
 
-    **Example Request**:
-
     :param unitid: The UUID of the organisational unit.
```
cmoesgaard: Update the docstring of this function to make it clear that we are now editing the details for an org unit, and no longer just the org unit itself.

mora/service/orgunit.py
danchr: YAGNI — wrapping is for compatibility, and this function is new?

mora/service/orgunit.py
danchr: Indeed, and I'll fix the other docstring as well while at it!

mora/service/orgunit.py
cmoesgaard: I was merely thinking in terms of keeping our API consistent.
If the other instance of wrapping in a list is strictly an undocumented compatiblity thing, then let's just keep this as it is.

PR #73, opened 2018-03-20
#21861 itsystem validation

danchr:


mora/service/common.py
```diff
@@ -93,6 +93,37 @@ def get_connector():
     return lora.Connector(**loraparams)
 
 
+def checked_get(
+    mapping: dict,
+    key: Hashable,
+    default: Hashable,
```
cmoesgaard: I think calling this `default` is misleading, if we don't actually return it as a default value. Maybe this should just be a type instead?

mora/service/itsystem.py
```diff
@@ -223,19 +228,33 @@ def get_relation_for(value, start, end):
...
 
     def create(self, id, req):
+        systemobj = common.checked_get(req, keys.ITSYSTEM, {})
+        systemid = common.get_uuid(systemobj)
+
+        if not systemid:
+            raise ValueError('missing system uuid!')
+
+        if not util.is_uuid(systemid):
+            raise ValueError('invalid system uuid!')
```
cmoesgaard: I don't think we'll ever reach this branch... All the checks happen in the `get_uuid`/`checked_get` calls.

mora/service/itsystem.py
```diff
@@ -223,19 +228,33 @@ def get_relation_for(value, start, end):
...
-        rels.append(self.get_relation_for(
-            req[keys.ITSYSTEM]['uuid'],
-            common.get_valid_from(req),
-            common.get_valid_to(req),
-        ))
+        start = common.get_valid_from(req)
+        end = common.get_valid_to(req)
+
+        if start == util.negative_infinity:
+            raise ValueError('missing or invalid start date!')
```
cmoesgaard: We never want this to happen, right? Maybe build this into `get_valid_from`?

danchr: I actually considered that — I'll try to make the change and see if it breaks anything!

mora/service/itsystem.py
danchr: Hmmm, nice catch, I'll see if this code is dead, and drop it if so!

cmoesgaard: And the same goes for the check right above it, I think.

PR #77, opened 2018-04-04
Feature/21756 mox cpr

cmoesgaard:


config/mora-example.json
```diff
@@ -1,3 +1,10 @@
 {
-    "LORA_URL": "http://mox.example.com/"
+    "LORA_URL": "http://mox.example.com/",
+    "SP_SERVICE_UUIDS": {
+        "service_agreement": "deadbeef-dead-beef-dead-beefdeadbeef",
+        "user": "deadbeef-dead-beef-dead-beefdeadbeef",
+        "user_system": "deadbeef-dead-beef-dead-beefdeadbeef",
+        "service": "deadbeef-dead-beef-dead-beefdeadbeef"
```
danchr: `00000000-0000-0000-0000-000000000000`? Or perhaps `null`? Also, wouldn't it be nicer to have each of these as a separate setting?

On a totally unrelated note, I really want this to die — we should have used Flasks configuration support instead…

mora/integrations/serviceplatformen.py
```diff
@@ -0,0 +1,172 @@
...
+
+import random
+
+from requests import HTTPError
+
+from mora import settings
+
+
+def get_citizen(cpr):
+    if settings.PROD_MODE:
```
danchr: Shouldn't this appear in `mora-example.json`?

mora/integrations/serviceplatformen.py
```diff
@@ -0,0 +1,172 @@
...
+
+    if (int(cpr[-1]) % 2) == 0:
+        first_name = random.choice(FEMALE_FIRST_NAMES)
+    else:
+        first_name = random.choice(MALE_FIRST_NAMES)
+
+    return {
+        'fornavn': first_name,
+        'efternavn': random.choice(LAST_NAMES)
+    }
```
danchr: Shouldn't we make this consistent rather than random? For testing?

mora/service/common.py
```diff
@@ -523,6 +523,53 @@ def create_organisationsenhed_payload(
...
+    brugernavn: str,
+    brugervendtnoegle: str,
+    tilhoerer: str,
+    cpr: str
+):
+
+    virkning = _create_virkning(valid_from, valid_to)
+
+    user = {
+        'note': 'Oprettet i MO',
```
danchr: `Importeret fra ServicePlatformen`?

mora/service/cpr.py
```diff
@@ -0,0 +1,45 @@
...
+import flask
+
+from mora.integrations.serviceplatformen import get_citizen
+from mora.service import keys
+
+blueprint = flask.Blueprint('cpr', __name__, static_url_path='',
+                            url_prefix='/service')
+
+
+@blueprint.route('/e/cpr_lookup/<string:cpr>/')
```
danchr: This feels vaguely inconsistent — perhaps something like `/e/create` or `/e/import`?

mora/service/cpr.py
```diff
@@ -0,0 +1,45 @@
...
+    return flask.jsonify(format_cpr_response(sp_data, cpr))
+
+
+def format_cpr_response(sp_data: dict, cpr: str):
+    first_name = sp_data.get('fornavn')
+    middle_name = sp_data.get('mellemnavn')
+    last_name = sp_data.get('efternavn')
+
+    # Filter empty name components, and construct full name string
+    name = ' '.join(filter(lambda x: x, [first_name, middle_name, last_name]))
```
danchr: Using `filter` and `map` with an inline lambda is an anti-pattern — either use `filter(None, …)`, a generator or a list comprehension.

mora/service/employee.py
```diff
@@ -189,10 +191,10 @@ def get_employee(id):
 
 
 @blueprint.route('/e/<uuid:employee_uuid>/create', methods=['POST'])
-def create_employee(employee_uuid):
+def create_employee_relation(employee_uuid):
     """Creates new employee relations
 
-    .. :quickref: Employee; Create
+    .. :quickref: Employee; Create relation
```
danchr: What should probably elaborate here — a relation to what? The moon?

tests/test_integration_employee.py
```diff
@@ -0,0 +1,139 @@
...
+
+from mora import lora
+from . import util
+
+mock_uuid = 'f494ad89-039d-478e-91f2-a63566554bd6'
+
+
+@freezegun.freeze_time('2017-01-01', tz_offset=1)
+@patch('mora.service.employee.uuid.uuid4', new=lambda: mock_uuid)
+class Tests(util.LoRATestCase):
```
danchr: Note: you're not actually testing the `serviceplatformen` module at all.

tests/test_integration_employee.py
```diff
@@ -0,0 +1,139 @@
...
+import freezegun
+
+from mora import lora
+from . import util
+
+mock_uuid = 'f494ad89-039d-478e-91f2-a63566554bd6'
+
+
+@freezegun.freeze_time('2017-01-01', tz_offset=1)
+@patch('mora.service.employee.uuid.uuid4', new=lambda: mock_uuid)
```
danchr: Mock `uuid.uuid4` instead — it's the same object, and if that doesn't work, someone is doing something wrong. Also, it's more concise and readable.

config/mora-example.json
cmoesgaard: The service UUID structure is what the SP library expects as a parameter. It does make more sense to split it up, though. I'll do that.

mora/integrations/serviceplatformen.py
cmoesgaard: It is consistent, `random` is seeded with `cpr`

mora/service/cpr.py
cmoesgaard: This isn't a creation or import, but rather just a lookup in SP on a CPR. The resulting data is not written to LoRa, but returned to the UI.

tests/test_integration_employee.py
cmoesgaard: I honestly think it makes more sense to mock out the specific use of `uuid`, rather than mocking every use of it everywhere.

mora/integrations/serviceplatformen.py
cmoesgaard: It should. I'll fix it.

mora/service/employee.py
cmoesgaard: I tried to be consistent with what is described for the corresponding `/ou/` endpoint.
```
Creates new unit relations

    .. :quickref: Unit; Create relation
```

mora/integrations/serviceplatformen.py
danchr: Oh, right — I missed that line. Could you perhaps add a comment to highlight that above the `seed` call?

mora/service/employee.py
danchr: Right, it's fine then :)

mora/service/cpr.py
```diff
@@ -5,6 +5,12 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 #
+'''CPR
+-----------
+
+This section describes functionality for retrieving information about people
+based on CPR..
```
danchr: Did you finish this sentence too soon?

tests/test_integration_employee.py
```diff
@@ -9,14 +9,15 @@
...
+from mora import lora, settings
+from mora.service import employee, orgunit
 from . import util
 
 mock_uuid = 'f494ad89-039d-478e-91f2-a63566554bd6'
 
 
 @freezegun.freeze_time('2017-01-01', tz_offset=1)
-@patch('mora.service.employee.uuid.uuid4', new=lambda: mock_uuid)
+@patch('mora.service.employee.uuid4', new=lambda: mock_uuid)
```
danchr: `s/mora.service.employee.uuid4/uuid.uuid4/`

Since we're doing this all over the place, we should probably write a decorator that injects a simple, incrementing replacement for `uuid4()`.

tests/test_integration_employee.py
```diff
@@ -110,20 +111,14 @@ def test_create_employee(self):
...
-    def test_cpr_lookup(self, mock_get_citizen):
+    def test_cpr_lookup_prod_mode_false(self):
         # Arrange
-
-        mock_get_citizen.return_value = {
-            'fornavn': 'First',
-            'mellemnavn': 'Middle',
-            'efternavn': 'Last'
-        }
+        settings.PROD_MODE = False
```
danchr: You should either mock this or use `util.override_settings()` — sooner or later, you'll break some other test.

mora/service/cpr.py
danchr: Please clarify that in the docstring — along with documenting the response while at it.

We should probably convert the path parameter to a query parameter — that's more customary for searches — and please and make it use `util.restrictargs()` too.

mora/service/cpr.py
```diff
@@ -0,0 +1,51 @@
...
+@blueprint.route('/e/cpr_lookup/<string:cpr>/')
+def search_cpr(cpr):
+    """
+    Search for CPR in Serviceplatformen
+
+    :param cpr: The CPR no. of a person to be searched
+    """
+
+    if len(cpr) != 10:
+        raise ValueError('CPR should be 10 characters long')
```
danchr: Although this occurs many other places, I'd say the proper nomenclature is “CPR number”. Or _personal identification number_ if we want to be accurate and annoying 😉

mora/service/employee.py
```diff
@@ -14,6 +14,8 @@
 This section describes how to interact with employees.
 
 '''
+import datetime
+from uuid import uuid4
```
danchr: I thought we agreed on moving away from imports like these?

tests/test_integration_employee.py
cmoesgaard: My absolutely biggest gripe with replacing it with a more generic patching, as mentioned earlier, is that we then mock _every_ use of `uuid` everywhere. In this case it doesn't make a difference, as nothing else depends on it, or at least not in a way that breaks stuff when we mock it out, but we never know.

danchr: My point is that we _should_ patch `uuid4()` — `https://github.com/nicwest/notsouid` looks interesting…

cmoesgaard: I initially just rolled my own, simplified decorator that does something similar to notsouid. I'll just replace it and be done with it.

mora/service/common.py
cmoesgaard: As this code is triggered by an action in MO, and not an import from SP, I feel this is fine as it is.

PR #79, opened 2018-04-09
replace custom manage.py with pipenv configuration and a shell script

danchr:
This is an example on how to use [pipenv](https://docs.pipenv.org) for MO. Is this acceptable from a deployment perspective? The main advantage is that it makes it much easier to specify dependencies for development — _X or later_ — and deployment — _exactly X_.


danchr:
> It makes absolutely sense to simply the workflow for the developers.
>
> However in a deployment perspective, the original requirements.txt (for production) should be included for both compatibility and legacy purposes.

Personally, I'd rather we opt for one technology and use that whenever possible. We can generate a `requirements.txt` file if needs be, or perhaps add a command to generate `requirements{,-dev}.txt` from the `Pipfile`. Having two manually maintained files seems like something that'll inevitably cause pain…

PR #85, opened 2018-04-17
Use a separate key for addresses depending on the type

danchr:
This is a request from our frontenders.


mora/service/address.py
```diff
@@ -267,18 +270,23 @@ def edit(self, id, req):
...
 
         new_rel = get_relation_for(
             common.checked_get(new_entry, keys.ADDRESS_TYPE, {},
                                fallback=old_entry, required=True),
-            common.checked_get(new_entry, keys.VALUE, '',
-                               fallback=old_entry, required=True),
+            new_entry,
+            old_entry,
         )
+        print('NEW', new_rel)
```
cmoesgaard: I'm guessing you forgot to remove these.

mora/service/orgunit.py
```diff
@@ -146,7 +146,7 @@ def edit(self, unitid, req):
                     addrs,
                     address.get_relation_for(
                         original_data[keys.ADDRESS_TYPE],
-                        original_data[keys.VALUE],
+                        original_data,
                         original_data[keys.VALIDITY],
```
cmoesgaard: Am I tired, or are you sending the validity as the fallback here? Does that make sense?

mora/service/address.py
```diff
@@ -51,24 +51,26 @@
...
+        if (
+            'urn' in addrobj or
+            keys.VALUE not in addrobj and fallback and 'urn' in fallback
+        ):
+            return {
+                'urn': common.get_urn(addrobj, fallback),
+                'objekttype': classobj['uuid'],
+            }
+
+        value = common.checked_get(addrobj, keys.VALUE, '', required=True)
```
cmoesgaard: Shouldn't we use fallback here as well?

mora/service/orgunit.py
danchr: > Am I tired, or are you sending the validity as the fallback here? Does that make sense?

Erm, good question…

mora/service/address.py
danchr: > Shouldn't we use fallback here as well?

The line above constitutes the fallback — I'll add a comment to that effect.

mora/service/orgunit.py
danchr: >> Am I tired, or are you sending the validity as the fallback here? Does that make sense?

> Erm, good question…

The answer appears to be that our tests [do not cover](http://lorajenkins.atlas.magenta.dk/job/mora/view/change-requests/job/PR-85/1/cobertura/service/orgunit_py/) this code, so it couldn't possibly work 😞

mora/service/address.py
```diff
@@ -51,24 +51,31 @@
                             url_prefix='/service')
 
 
-def get_relation_for(classobj, value):
-    scope = classobj['scope']
-
-    if not isinstance(value, str):
-        value = str(value)
+def get_relation_for(classobj, addrobj, fallback=None):
```
cmoesgaard: Quasi-unrelated to the PR, but `s/classobj/addrtypeobj/`? Just to help make this function more easily parsable the next time I look at it.

mora/service/address.py
cmoesgaard: Right, having read the comment and the function one more time, I think it makes sense. This function could really use a refactoring or some more comments. Is the structure and the logic of the address object documented anywhere? Maybe we should refer to that from here...

PR #87, opened 2018-04-18
address: handle missing types

danchr:
I ran into this on moradev.

### DIFF
--- a/mora/service/address.py
+++ b/mora/service/address.py
@@ -333,12 +333,14 @@ class Addresses(common.AbstractRelationDetail):
                 if not c.is_effect_relevant(addrrel['virkning']):
                     continue
 
-                yield {
-                    **get_one_address(c, addrrel, class_cache),
-                    keys.ADDRESS_TYPE: get_address_class(c, addrrel,
-                                                         class_cache),
-                    keys.VALIDITY: common.get_effect_validity(addrrel),
-                }
+                try:
+                    addr = get_one_address(c, addrrel, class_cache)
+                except KeyError as e:
+                    continue
+
+                addr[keys.VALIDITY] = common.get_effect_validity(addrrel)
+
+                yield addr
 
         return flask.jsonify(
             sorted(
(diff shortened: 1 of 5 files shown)


mora/service/address.py
```diff
@@ -333,12 +333,14 @@ def convert(start, end, effect):
...
 
-                yield {
-                    **get_one_address(c, addrrel, class_cache),
-                    keys.ADDRESS_TYPE: get_address_class(c, addrrel,
-                                                         class_cache),
-                    keys.VALIDITY: common.get_effect_validity(addrrel),
-                }
+                try:
+                    addr = get_one_address(c, addrrel, class_cache)
+                except KeyError as e:
```
cmoesgaard: Where is KeyError thrown? Are we sure we're not swallowing actual exceptions here?

danchr: > Where is KeyError thrown? Are we sure we're not swallowing actual exceptions here?

In [`get_one_class()`](https://github.com/magenta-aps/mora/pull/87/commits/6d113c1ac5bbf45ba28ef130fd13fcb3c2cf9874#diff-8426190c255ed9b2d05701905cef41c1R132) — we _could_ also return `None` there and handle that, but an exception seemed more appropriate for what ought to be an exceptional case…

mora/util.py
```diff
@@ -289,3 +289,24 @@ def uniqueify(xs):
     '''return the contents of xs as a list, but stable'''
     # TODO: is this fast?
     return list(collections.OrderedDict(itertools.zip_longest(xs, ())).keys())
+
+
+def log_exception(msg='', with_traceback=False):
```
cmoesgaard: `with_traceback` doesn't seem to be used.

mora/service/address.py
cmoesgaard: As far as I can tell from this, the UI is unaware that something went wrong. Should they be aware? Or do we just expect whoever is in charge to check the log ever so often for errors?

mora/service/address.py
```diff
@@ -6,6 +6,10 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 #
 
+import json
+import sys
+import traceback
```
cmoesgaard: `sys` and `traceback` aren't used.

PR #88, opened 2018-04-25
#22274 orgunit address dates

danchr:
This fixes setting a different validity date on addresses when creating organisational units. The PR also includes another minor unrelated fix to logging I stumbled upon, and a change so that it's less likely that we target imports at the wrong location.

### DIFF
--- a/mora/cli.py
+++ b/mora/cli.py
@@ -325,8 +325,8 @@ def load_cli(app):
 
     @app.cli.command('import')
     @click.argument('spreadsheets', nargs=-1, type=click.Path())
-    @click.option('--destination-url', '-d',
-                  help='LoRA url')
+    @click.option('--dry-run', '-n', is_flag=True,
+                  help=("don't actually change anything"))
     @click.option('--verbose', '-v', count=True,
                   help='Show more output.')
     @click.option('--jobs', '-j', default=1, type=int,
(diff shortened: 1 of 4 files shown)


mora/service/orgunit.py
```diff
@@ -446,7 +446,7 @@ def create_org_unit():
     org_uuid = organisationenhed_get['relationer']['tilhoerer'][0]['uuid']
     org_unit_type_uuid = req.get(keys.ORG_UNIT_TYPE).get('uuid')
     addresses = [
-        address.get_relation_for(addr)
+        address.get_relation_for(addr, req)
```
cmoesgaard: Why is `req` used as fallback here? Is it to get the validity from the request?

danchr:
> Why is `req` used as fallback here? Is it to get the validity from the request?

Correct.

mora/service/orgunit.py
danchr: > Why is `req` used as fallback here? Is it to get the validity from the request?

Correct.

cmoesgaard: I am really not a fan of this. So far, the `fallback` has been used to supply alternate address objects to the function as a fallback, so to now use it for an arbitrary request object that just so happens to have some properties we need, seems like an abuse of the parameter more than anything.

danchr: I get what you mean, but the use is not _entirely_ inconsistent: Whereas we'd previously get just address & type from the object itself, now we also get address, type and validity.

The main downside, of course, is that it might allow some truly weird requests to succeed in an unexpected manner, where the value gets picked up from the request itself rather than the addresses.

Coming to think of it, this code might not be necessary; I'll check…

PR #92, opened 2018-05-01
#22290 adjust user objects to be valid from their birthday

danchr:


mora/util.py
```diff
@@ -310,3 +310,36 @@ def log_exception(msg=''):
...
+    elif code < 9000:
+        century = 2000 if year <= 57 else 1800
+    else:
+        century = 2000 if year <= 36 else 1900
+
+    try:
+        return (
+            datetime.datetime(century + year, month, day,
+                              tzinfo=default_timezone),
+            code,
```
cmoesgaard: Why are we returning the code?

danchr: > Why are we returning the code?

My idea was that this function simply parsed the CPR number into its two main constituent parts. We're not using it anywhere, though, so I guess _YAGNI_ applies. I'll remove it…

PR #93, opened 2018-05-03
cpr test: only return something for valid CPR numbers

danchr:
By 'valid' we mean the ones that represent a valid date of birth.

### DIFF
--- a/mora/integrations/serviceplatformen.py
+++ b/mora/integrations/serviceplatformen.py
@@ -15,6 +15,9 @@ from .. import settings
 
 
 def get_citizen(cpr):
+    if not util.is_cpr_number(cpr):
+        raise ValueError('invalid CPR number!')
+
     if settings.PROD_MODE:
         try:
             # TBC when SP library is extracted from AVA repo
@@ -158,9 +161,6 @@ LAST_NAMES = [
 
 
 def _get_citizen_stub(cpr):
-    if not util.is_cpr_number(cpr):
-        return None
-
     # Seed random with CPR number to ensure consistent output
     random.seed(cpr)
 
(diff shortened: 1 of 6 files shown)


mora/integrations/serviceplatformen.py
```diff
@@ -157,6 +158,9 @@ def get_citizen(cpr):
 
 
 def _get_citizen_stub(cpr):
+    if not util.is_cpr_number(cpr):
```
cmoesgaard: Move this to `get_citizen` so we do this check regardless of running the stub or not.

mora/service/cpr.py
```diff
@@ -48,12 +48,20 @@ def search_cpr():
...
-    else:
-        return '', 404
+    try:
+        sp_data = get_citizen(cpr)
+    except KeyError:
+        raise exceptions.NotFoundError(
+            'no such person found',
+            cpr=cpr,
+        )
+    except ValueError:
```
cmoesgaard: Why catch and re-raise these exceptions? Wouldn't it make more sense to just raise the relevant errors in `get_citizen` directly?

danchr: Layering? Figured the `integrations` package wasn't as MO-specific, and should raise plain exceptions, leaving the interpretation to the `cpr_lookup` service.

cmoesgaard: I did intend it as something MO-specific, actually. The `integrations` package is to contain all of the adapters and logic for connecting to external systems - but I see your point. This way we could potentially extract this and reuse it somewhere else.

danchr: Anyway, my thinking was that `KeyError` and `ValueError` talk of arguments to the current function, whereas the `exceptions` module is for shortcutting evaluation with an HTTP error — and that felt like something an integration shouldn't do.

PR #95, opened 2018-05-08
Feature/22084 validation

cmoesgaard:
Validations for ensuring that the creation of org funk objects are inside of the ranges of their associated employee and org unit objects.
Implementation of error codes for error messages sent to the UI.

There are still many validations that haven't been implemented yet, but I'd rather get this merged in now, to prevent having a giant merge later on.


mora/errors.py
```diff
@@ -0,0 +1,42 @@
...
+    E37 = "No local municipality found."
+    E38 = "Size must be positive."
+
+    # Misc
+    E90000 = "Invalid input."
+    E90001 = "Unauthorized."
+    E90002 = "Connection failed."
+    E90003 = "Not found."
+    E90004 = "No such endpoint."
+    E99999 = "Unknown Error."
```
danchr: I'm not fond of the use of numeric identifiers for the enum values; they're make little or no sense when reading the code. Can't we use `E_UNKNOWN_ERROR` or `ERR_UNKNOWN` or something similar? Also, I see no requirement that errors codes be numeric; both Python and JavaScript handle strings quite well.

Also, the name of this file clashes conceptually with `exceptions.py`, as the two words are mostly synonymous. Perhaps `errorcodes.py` instead?

mora/service/address.py
```diff
@@ -415,7 +416,7 @@ def edit(self, id, req):
 
         if not old_entry:
             raise exceptions.ValidationError(
-                'original required!',
+                Error.V4
```
danchr: This is an example of how the code is _less_ readable now. Previously, you'd know what the error was from this location alone — now you have to inspect the value in `errors.py`.

mora/app.py
```diff
@@ -45,7 +45,7 @@ def handle_invalid_usage(error):
 
     return exceptions.BaseError(
         description=str(error),
-        key=Error.E99999,
+        key=ErrorCodes.E_UNKNOWN_ERROR,
```
danchr: Minor nit: if E means error, no need to say it again.

mora/exceptions.py
```diff
@@ -49,18 +49,18 @@ def get_body(self, environ=None):
...
+    key = ErrorCodes.E_INVALID_INPUT
     cause = 'validation'
     code = 400
 
 
 class NotFoundError(BaseError):
-    key = Error.E90003
+    key = ErrorCodes.E_NOT_FOUND
     cause = 'not-found'
     code = 404
```
danchr: I'd like us to get rid of these exception, and encode the `code` somewhere else — such as in your prefix character.

cmoesgaard: So you want it to be `E404_NOT_FOUND`?

danchr: Or perhaps:

```python
E_NOT_FOUND = 404, 'go fetch'
```

cmoesgaard: I guess that makes sense. So we would only have a single error class, essentially, and have the individual codes be defined in `errorcodes`?

mora/app.py
```diff
@@ -52,6 +54,6 @@ def handle_invalid_usage(error):
 @app.route('/<path:path>')
 def root(path=''):
     if path.split('/', 1)[0] == 'service':
-        raise exceptions.NotFoundError('no such endpoint')
+        raise exceptions.NotFoundError(ErrorCodes.E_NO_SUCH_ENDPOINT)
```
danchr: …as an example, we now have some duplication between the exception object and the enum here — and the exception doesn't really add anything.

mora/lora.py
```diff
@@ -10,12 +10,13 @@
...
 import collections
 import datetime
-import itertools
 import functools
+import itertools
 import uuid
 
 import requests
 
+from mora.exceptions import ErrorCodes
```
danchr:
```python
from .exceptions import ErrorCodes```

File: mora/exceptions.py [hunk 7]
```diff
@@ -7,55 +7,81 @@
...
+    # Misc
+    E_INVALID_INPUT = 400, "Invalid input."
+    E_UNAUTHORIZED = 401, "Unauthorized."
+    E_CONNECTION_FAILED = 500, "Connection failed."
+    E_NOT_FOUND = 404, "Not found."
+    E_NO_SUCH_ENDPOINT = 404, "No such endpoint."
+    E_UNKNOWN = 500, "Unknown Error."
 
 
 class BaseError(werkzeug.exceptions.HTTPException):
```
danchr: We should rename this class now that it no longer has any subclass, and hence isn't a base error — perhaps just call it `HTTPException` to signify its relation to `werkzeug`?

mora/service/association.py
```diff
@@ -16,7 +16,7 @@
 
 import flask
 
-from mora import lora
+from mora import lora, validator
```
danchr: `from .`

mora/service/details.py
```diff
@@ -21,6 +21,7 @@
 
 import flask
 
+from mora.exceptions import ErrorCodes
```
danchr: `from .exceptions`

mora/service/employee.py
```diff
@@ -20,6 +20,8 @@
 import flask
 import werkzeug
 
+from .. import exceptions
+from mora.exceptions import ErrorCodes
```
danchr: You can delete the latter import — the former is sufficient.

mora/service/engagement.py
```diff
@@ -14,7 +14,7 @@
 employees and organisational units.
 
 '''
-
+from mora import validator
```
danchr: `from . import`

mora/service/itsystem.py
```diff
@@ -21,6 +21,7 @@
 import flask
 import werkzeug
 
+from mora.exceptions import ErrorCodes
```
danchr: `from . import`

mora/service/mapping.py
```diff
@@ -5,7 +5,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 #
-
+from mora.service import common, keys
```
danchr: These could be relative imports — and you've duplicated the import of `common`. (Also, see [PEP8](https://www.python.org/dev/peps/pep-0008/#imports) on imports — modules should be one per line.)

mora/service/orgunit.py
```diff
@@ -22,6 +22,7 @@
 import flask
 import werkzeug
 
+from mora.exceptions import ErrorCodes
```
danchr: I _really_ like relative imports.

mora/service/orgunit.py
```diff
@@ -810,6 +807,9 @@ def get_org_unit_history(unitid):
                                                  registreretfra='-infinity',
                                                  registrerettil='infinity')
 
+    if not unit_registrations:
+        raise werkzeug.exceptions.NotFound('no such unit')
```
danchr: Needs test [coverage](http://lorajenkins.atlas.magenta.dk/job/mora/view/change-requests/job/PR-95/13/cobertura/service/orgunit_py/)?

mora/service/role.py
```diff
@@ -16,7 +16,7 @@
 
 import flask
 
-from mora import lora
+from mora import lora, validator
```
danchr: `from ..` on separate lines?

mora/tokens.py
```diff
@@ -14,6 +14,7 @@
 import flask
 import requests
 
+from mora.exceptions import ErrorCodes
```
danchr: Do we really need to import both the module and the enum?

tests/test_service_common.py
```diff
@@ -1939,23 +1939,23 @@ def test_checked_get(self):
...
+                exceptions.BaseError,
+                'Invalid type',
         ):
             common.checked_get(mapping, 'dict', [])
 
         with self.assertRaisesRegex(
-                exceptions.ValidationError,
-                r"invalid 'list', expected dict, got: \[1337\]",
+                exceptions.BaseError,
+                r"Invalid type",
```
danchr: Can we extend this with a test that verifies that the output is helpful? For example, that it includes the name of the failing key, and preferably the invalid value too. (We should help the client by informing them _how_ they're wrong, rather than merely passing blame.)

PR #99, opened 2018-05-15
Feature/20648 auth implementation

istherepie:
Added a before_request method to fetch the authorization token from a HTTP request (if any) and attach it to the flask application context. 

Extended requests.Session() to subsequently pass on the authorization header/token from the user request to the backend (LORA).

### DIFF
--- a/mora/lora.py
+++ b/mora/lora.py
@@ -15,18 +15,13 @@ import functools
 import uuid
 
 import requests
+from flask import g
 
 from . import auth
 from . import exceptions
 from . import settings
 from . import util
 
-session = requests.Session()
-session.verify = settings.CA_BUNDLE or True
-session.auth = auth.SAMLAuth()
-session.headers = {
-    'User-Agent': 'MORA/0.1',
-}
 
 ALL_RELATION_NAMES = {
     'adresse',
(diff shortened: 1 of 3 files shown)


mora/lora.py
```diff
@@ -145,6 +140,66 @@
...
+        and attaching the token value to the application context.
+
+        :param context:     Application context
+                            For more information,
+                            head to the official flask documentation.
+        """
+
+        # Checks
+        if not hasattr(context, "user_token"):
+            raise AttributeError(
```
danchr: I believe this is wrong; we delegate authentication to LoRA, and should only reject incoming requests if LoRA does so as well. We should reject such requests by communicating the appropriate HTTP status code to the browser, using the infrastructure in `mora/exceptions.py`.

Furthermore, given the requirements for a fully authenticated setup, we should only fail if authentication is in fact enabled and the request fails. Otherwise, we'll break most of our current installations.

mora/lora.py
```diff
@@ -145,6 +140,66 @@
...
+                'Missing user_token attribute'
+            )
+
+        # TODO: Replace exception with proper exception type
+        if not context.user_token:
+            raise ValueError(
```
danchr: The proper type would be `HTTPException` from `mora/exceptions.py`, but this is not the right place to raise it, as earlier mentioned.

tests/test_authorized_session.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+# -- coding: utf-8 --
+#
+# Copyright (c) 2017-2018, Magenta ApS
+#
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
```
danchr: Please remove the duplicate license header.

tests/test_authorized_session.py
```diff
@@ -0,0 +1,75 @@
...
+
+        with app.test_client() as client:
+
+            # GET request
+            # No auth header
+            client.get(path='/service/e/create')
+
+            # Should fail due to missing token in context
+            # TODO: ValueError should be replaced with custom exception
+            self.assertRaises(ValueError, AuthorizedSession, context=g)
```
danchr: This test is wrong, as it verifies that we reject all unauthenticated requests — which we shouldn't.

danchr:
This one has been languishing for a week or so, so let's close it — we can reopen it later, if needs be.

PR #101, opened 2018-05-16
Feature/22084 more validations

cmoesgaard:


mora/service/leave.py
```diff
@@ -35,8 +35,7 @@ def create_leave(employee_uuid, req):
         employee_uuid)['relationer']['tilhoerer'][0]['uuid']
     leave_type_uuid = common.get_mapping_uuid(req, keys.LEAVE_TYPE,
                                               required=True)
-    valid_from = common.get_valid_from(req)
-    valid_to = common.get_valid_to(req)
+    valid_from, valid_to = common.get_validities(req)
```
danchr: Neat — nice and Pythonic!

mora/service/orgunit.py
```diff
@@ -135,12 +138,12 @@ def edit(self, unitid, req):
             ))
 
         if keys.PARENT in data.keys():
```
danchr: Although not relevant to this change, I'd say this is an anti-pattern. `if keys.PARENT in data:` is the appropriate syntax…

mora/validator.py
```diff
@@ -139,52 +141,75 @@ def is_node_valid(node_uuid: str) -> bool:
...
+        c.organisationenhed.get_effects(
+            unitid,
+            {
+                'tilstande': (
+                    'organisationenhedgyldighed',
+                ),
+            },
+            {}
+        )
+        if effect.get('tilstande')
```
danchr: Lose the uses of `.get()` here — it's a lazy waste of characters 😉 If you don't handle a the default value, use simple item access.

tests/test_integration_association.py
```diff
@@ -893,6 +942,82 @@ def test_edit_association_no_overwrite(self):
...
+                },
+                "validity": {
+                    "from": "2017-12-01T00:00:00+01",
+                    "to": "2017-12-02T00:00:00+01",
+                },
+            }
+        ]
+
+        self._perform_request('/service/e/{}/create'.format(userid),
+                              json=payload)
```
danchr: Shouldn't we check the result of this invocation? Alternatively, we should perhaps change the result be useful here.

tests/test_integration_org_unit.py
```diff
@@ -397,6 +397,63 @@ def test_create_org_unit(self, m):
...
+        }
+
+        expected = {
+            'description': 'Date range exceeds validity '
+                           'range of associated org unit.',
+            'error': True,
+            'error_key': 'V_DATE_OUTSIDE_ORG_UNIT_RANGE',
+            'org_unit_uuid': '2874e1dc-85e6-4269-823a-e1125484dfd3',
+            'status': 400,
+            'valid_from': 'Thu, 04 Feb 2010 00:00:00 GMT',
```
danchr: This also isn't related to this change, strictly speaking, but we should fix this to use ISO 8601 format just like the rest of our API.

tests/test_integration_org_unit.py
```diff
@@ -847,28 +904,28 @@ def test_rename_org_unit_early(self):
...
                 'parent': {
-                    'name': 'Humanistisk fakultet',
-                    'user_key': 'hum',
-                    'uuid': '9d07123e-47ac-4a9a-88c8-da82e3a4bc9e',
+                    'name': 'Overordnet Enhed',
+                    'user_key': 'root',
+                    'uuid': '2874e1dc-85e6-4269-823a-e1125484dfd3',
                 },
-                'user_key': 'fil',
+                'user_key': 'samf',
```
danchr: Shouldn't we, eventually, validate the unit keys are unique? If so, this'll conflict with the preexisting unit.

tests/test_integration_org_unit.py
```diff
@@ -1044,6 +1101,66 @@ def test_move_org_unit_should_fail_validation(self):
             status_code=400,
             json=req)
 
+    def test_move_org_unit_should_fail_when_moving_root_unit(self):
+        """Should fail validation when trying to move the root org unit"""
```
danchr: Indeed it should!

mora/service/orgunit.py
cmoesgaard: You've mentioned that before. I'll run through them all and fix it.

danchr: Thanks :)

tests/test_integration_org_unit.py
cmoesgaard: We should, eventually. The unit keys are randomly generated for now, with no way for the user to affect it, so I think we should cross this bridge when we get to it.

PR #103, opened 2018-05-22
Fixed duplication of addresses on org units.

cmoesgaard:
Simplified presentation of phone numbers to better conform with what the
UI expects.


mora/service/address.py
```diff
@@ -360,19 +362,7 @@ def convert(start, end, effect):
...
-                            id,
-                            {
-                                'relationer': (
-                                    'adresser',
-                                ),
-                            }
-                        ),
-                    ),
-                ),
+                convert(self.scope.get(id)),
```
danchr: Won't this break past/current/future support for addresses?

PR #104, opened 2018-05-22
Feature/internal lora tests

cmoesgaard:


tests/util.py
```diff
@@ -397,141 +405,127 @@ def sort_inner_lists(obj):
...
+        oio_rest.app.app.config["DEBUG"] = True
+        oio_rest.app.app.config["TESTING"] = True
+        oio_rest.app.app.config['PRESERVE_CONTEXT_ON_EXCEPTION'] = False
 
-        for t in map(c.__getattr__, c.scope_map):
-            for objid in t(bvn='%'):
-                t.delete(objid)
+        if hasattr(db.adapt, 'connection'):
+            del db.adapt.connection
+            self.addCleanup(lambda: db.adapt.connection.close())
```
cmoesgaard: Shouldn't we close the connection regardless of whether it was opened before? With this, the cleanup function is not going to be registered on the first test run.

danchr: > Shouldn't we close the connection regardless of whether it was opened before? With this, the cleanup function is not going to be registered on the first test run.

Erm, this change is overwritten by https://github.com/magenta-aps/mora/pull/104/commits/86cca32e1262fac1f3f587ec712bf9f03c25cd67?

cmoesgaard: ...I'm tired.

cmoesgaard:
Having recently rediscovered how the ordering of the pull request commit history works, I think this looks fine overall.

danchr:
> Having recently rediscovered how the ordering of the pull request commit history works, I think this looks fine overall.

Yeah, GitHub orders commits [by date](https://help.github.com/articles/why-are-my-commits-in-the-wrong-order/) rather than the proper, true and sensible topological order… 🙄

PR #105, opened 2018-05-28
Feature/20648 auth implementation

istherepie:
Minor changes to auth handler. 
Removed username from URI request path
Username is passed as part of the json payload

### DIFF
--- a/mora/auth.py
+++ b/mora/auth.py
@@ -33,8 +33,8 @@ COOKIE_NAME = 'MO-Token'
 
 basedir = os.path.dirname(__file__)
 
-blueprint = flask.Blueprint('authentication', __name__, url_prefix='/mo',
-                            root_path=basedir)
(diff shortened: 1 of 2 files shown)


mora/auth.py
```diff
@@ -108,13 +111,12 @@ def login(username):
     return resp
 
 
-@blueprint.route('/service/user/<user>/logout', methods=['POST'])
+@blueprint.route('/service/user/logout', methods=['POST'])
 def logout(user=None):
```
danchr: Please remove the parameter as well.

istherepie:
Requested changes fixed.

danchr:
Oh, and it would be great if we could move the endpoint away from `/mo` — that's legacy cruft too. The blueprint prefix should just be `/service`.

istherepie:
Fixed as requested.

danchr:
Awesome, thanks!

PR #114, opened 2018-06-12
#23054 employee engagements atomicity

danchr:
This change makes creation of details atomic, such that either all of
them succeed or all fail. Within reason, that is: this only applies to
validation, and won't handle, say, LoRA dying in the middle of a
request gracefully.

### DIFF
--- a/mora/service/employee.py
+++ b/mora/service/employee.py
@@ -15,6 +15,7 @@ This section describes how to interact with employees.
 
 '''
 
+import collections
 import uuid
 
 import flask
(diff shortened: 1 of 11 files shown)


mora/service/employee.py
```diff
@@ -442,29 +443,61 @@ def create_employee_relation(employee_uuid):
...
+    note = "Opret {}".format(', '.join(types))
 
+    if not rels:
         # Write a noop entry to the user, to be used for the history
         common.add_bruger_history_entry(
             employee_uuid,
-            "Opret {}".format(common.RELATION_TRANSLATIONS[role_type])
+            note,
         )
+    else:
```
cmoesgaard: This kind of goes against the original intent of this function.

The idea was to have a relatively generic function, with a number of handlers sharing the same interface, keeping all of the specific nitty-gritty of creating the objects and calling LoRa in the specific handlers, rather than here.

I'd rather we modify the handlers for this to basically implement two functions -- one two create a payload, and one to send the payload to LoRa, and try to unify the two different types of handlers we have created here. That way, we still perform all the necessary validation for every object before updating the database, while keeping this function relatively simple.

Handling history in a elegant way _is_ a bit difficult, though, I realize.

PR #117, opened 2018-06-18
include validity when returning organisational units

danchr:
We generally only return the current validity, on the general
assumption that this will span a sufficiently long interval.

This change supersedes #114.

### DIFF
--- a/mora/service/orgunit.py
+++ b/mora/service/orgunit.py
@@ -184,6 +184,7 @@ def get_one_orgunit(c, unitid, unit=None,
 
     attrs = unit['attributter']['organisationenhedegenskaber'][0]
     rels = unit['relationer']
+    validities = unit['tilstande']['organisationenhedgyldighed']
 
     r = {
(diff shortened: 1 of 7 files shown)


mora/service/orgunit.py
```diff
@@ -223,8 +224,7 @@ def get_one_orgunit(c, unitid, unit=None,
             'invalid details {!r}'.format(details),
         )
 
-    if validity is not None:
-        r[keys.VALIDITY] = validity
+    r[keys.VALIDITY] = validity or common.get_effect_validity(validities[0])
```
cmoesgaard: Probably somewhat unrelated, but what is the point of the `validity` parameter to this function? Just to be used here?

danchr: > Probably somewhat unrelated, but what is the point of the `validity` parameter to this function? Just to be used here?

We use it for the `/service/ou/$UUID/details/org_unit` calls — if you remove it, those tests fail.

PR #119, opened 2018-06-22
Feature/21273 test fixes

cmoesgaard:
… now.

Updated tests.

Has a mutual dependency on a [corresponding branch of the mox repo](https://github.com/magenta-aps/mox/tree/feature/21273_searches_and_sorting), which needs to be merged to `development` at roughly the same time as this.

Jenkinsfile is set to point at that branch temporarily, to prove that the tests pass.


danchr:
In eafa813952b9e2ab9cfdffe1836fd5ae36ec3f8e, I changed it so that we exclude the 800 auto-generated users. Perhaps we should apply that one before this change to make it more manageable?

danchr:
On a related note, we might want to make this optional, if possible. Otherwise, we'll need to upgrade/replace `moxdev` and `moxdemo`:

```
+ ssh moradev cat /srv/mora/config/mora.json
{
    "USE_PATCH": false,
    "LORA_URL": "http://192.168.122.117/"
}
+ ssh morademo cat /srv/mora/config/mora.json
{
    "LORA_URL": "http://192.168.122.114/"
}
+ ssh moxdev cat /srv/mox/VERSION ; /srv/mox/python-env/bin/python --version
0.3.0.1
Python 2.7.12
+ ssh moxdemo cat /srv/mox/VERSION ; /srv/mox/python-env/bin/python --version
0.3.0.1
Python 2.7.12
```

cmoesgaard:
Good points - I'd rather we fix our environments than include more legacy handling. I'll see what I can do about that.

I'll go ahead and cherrypick eafa813 as well.

cmoesgaard:
`moxdev` has been updated with the newest stable LoRa, so this should be good to go :tm:

PR #124, opened 2018-07-02
#19505 TestCafe

danchr:
This series adds our very first end-to-end test using TestCafé. I had to change authentication a bit to ensure that it works as expected on a server without it configured.


README.rst
```diff
@@ -295,6 +305,14 @@ gøres på følgende måde::
   $ ./manage.py test --minimox=/path/to/folder/minimox tests.test_integration.IntegrationTests
   $ ./manage.py test --minimox=/path/to/folder/minimox tests.test_integration.IntegrationTests.test_should_add_one_new_contact_channel_correctly
 
+End-to-end tests
+----------------
+
+Vores end-to-end tests køres typisk som en del af testsuiten. For at
+køre den direkte mød en udviklingsmaskine anvendes eksempelvist::
```
cmoesgaard: s/mød/mod

danchr:
> The approach seems sound, but as far as I can see we use one LoRa instance for all tests. Isn't that going to be a problem in the long run in terms of keeping tests deterministic?

Yes, and I really want to fix that — but doing so is nontrivial, alas…

We have two options for this: a reporter plugin, or some form of callback. Both are tricky to write and/or get right, and I wanted to get this in first. From my perspective, the main downside is that we can't easily test different scenarios, such as single vs. multiple organisations, with this infrastructure. I'd like to fix that, eventually, but I think this'll do for now.

PR #125, opened 2018-07-03
#23172 proper termination error

danchr:
This PR makes the error message for termination lie less; previously,
it'd claim the date was too early even if it was, in fact, too late.
Instead, we just re-use the message saying that the date is outside
the validity of the unit. Since we were doing that, I consolidated the
termination check into the validity check, and added a frontend
validation instead, since they're generally nicer and more immediate.

### DIFF
--- a/mora/validator.py
+++ b/mora/validator.py
@@ -105,10 +105,14 @@ def _get_active_validity(reg: dict) -> typing.Mapping[str, str]:
 
 
 def is_date_range_in_org_unit_range(org_unit_uuid, valid_from, valid_to):
+    # query for the full range of effects; otherwise,
+    # _get_active_validity() won't return any useful data for time
+    # intervals predating the creation of the unit
     scope = lora.Connector(
         virkningfra=util.to_lora_time(util.NEGATIVE_INFINITY),
         virkningtil=util.to_lora_time(util.POSITIVE_INFINITY)
     ).organisationenhed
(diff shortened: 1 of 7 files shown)


mora/validator.py
```diff
@@ -104,11 +106,16 @@ def _get_active_validity(reg: dict) -> typing.Mapping[str, str]:
 
 def is_date_range_in_org_unit_range(org_unit_uuid, valid_from, valid_to):
     scope = lora.Connector(
-        virkningfra=util.to_lora_time(valid_from),
-        virkningtil=util.to_lora_time(valid_to)
+        virkningfra=util.to_lora_time(util.NEGATIVE_INFINITY),
+        virkningtil=util.to_lora_time(util.POSITIVE_INFINITY)
```
cmoesgaard: Why was this change made?

danchr: The trick is that the termination check can query a unit _prior_ to its creation — hence, merely obtaining the validity at the date of the check yields little useful data, and prevents us from returning the helpful _actual_ validity.

cmoesgaard: Ah, so this is purely to be able to return more helpful error messages?

In any case, I would like this function to be as similar to `is_date_range_in_employee_range` as possible, so it would be nice if similar adjustments could be made there as well - also in terms of the added not-found check

PR #131, opened 2018-08-07
Minor tweaks

danchr:
Minor tweaks made during or in preparation of the [AAK import](https://redmine.magenta-aps.dk/issues/22604). Tested using the test suite.


mora/service/orgunit.py
```diff
@@ -473,14 +473,25 @@ def create_org_unit():
...
+    if organisationenhed_get:
+        org_uuid = organisationenhed_get['relationer']['tilhoerer'][0]['uuid']
+    else:
+        organisation_get = c.organisation(uuid=parent_uuid)
+
+        if organisation_get:
+            org_uuid = parent_uuid
+        else:
+            raise exceptions.HTTPException(
+                exceptions.ErrorCodes.V_ORG_NOT_FOUND,
```
cmoesgaard: I don't think this error message is descriptive enough here, as we'll also reach this point if an org unit isn't found. Maybe a new error message showing that we did a lookup on `parent_uuid` in both `organisation` and `organisationenhed`?

cmoesgaard: Also, the docstring should be updated to reflect this.

mora/util.py
```diff
@@ -97,10 +99,16 @@ def parsedatetime(s: str) -> datetime.datetime:
...
-        )
+        if default is not _sentinel:
+            return default
+        else:
+            raise exceptions.HTTPException(
+                exceptions.ErrorCodes.E_INVALID_INPUT,
+                'cannot parse {!r}'.format(s)
+            )
+
+    if dt.date() == POSITIVE_INFINITY.date():
```
cmoesgaard: This seems like a change specific to how the data is for AAK.
I'd rather see us have the specific handling in the individual importers, than have a number of special cases here.

mora/util.py
```diff
@@ -97,10 +99,16 @@ def parsedatetime(s: str) -> datetime.datetime:
     try:
         dt = dateutil.parser.parse(s, dayfirst=True, tzinfos=_tzinfos)
     except ValueError:
-        raise exceptions.HTTPException(
-            exceptions.ErrorCodes.E_INVALID_INPUT,
-            'cannot parse {!r}'.format(s)
-        )
+        if default is not _sentinel:
```
cmoesgaard: Do we actually need a sentinel here? Is having `default` be `None` in the signature, and having `if default is not None` not enough?

mora/util.py
cmoesgaard: I see. I guess this is a general enough handling to be included in here.

PR #132, opened 2018-08-09
test_cafe: don't assume current working directory

danchr:
This is likely what broke the LoRA test suite — and since we're doing
this in a hotfix, bump the version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "mora",
-  "version": "0.7.1",
+  "version": "0.7.2",
   "description": "OS2 Mennesker + Organisation 2.0",
   "author": "Magenta ApS <info@magenta.dk>",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "mora",
-  "version": "0.7.1",
+  "version": "0.7.1.1",
```
cmoesgaard: Why not just 0.7.2?

danchr: It's a hotfix that doesn't change anything?

cmoesgaard: I mean, our versioning scheme so far as been major/minor/patch - Last hotfix bumped the version to 0.7.1 so this should be 0.7.2 in my opinion

danchr: Fair enough — fixed!

PR #134, opened 2018-08-10
Feature/22849 saml sso

cmoesgaard:
First run at implementing support for SAML SSO

We now both support SSO and the original SAML token based auth, all of which is configured and controlled in the middle-end

I've made a number of changes to the UI which might need special attention, notably moving our own login form to a separate page instead of a modal, as well as modifying the auth logic.

The build fails currently, due to the version of `pysaml2` potentially having a bug in its setup.py config in the newest version released a few days ago. (https://github.com/IdentityPython/pysaml2/issues/535)
So we're waiting for a resolution to that.


c00kiemon5ter:
Hello, and thanks for using `pysaml2`. There is no bug in `setup.py`/`setup.cfg`. Please, upgrade your environment to use the latest stable version of `setuptools` (Current version: `40.0.0`.)

e2e-tests/LoginTest.js
```diff
@@ -6,32 +6,18 @@ fixture('Login')
 
 test('Test empty login', async t => {
   await t
-    .click(Selector('.btn-text').withText('Organisation'))
-    .expect(Selector('#login.modal').exists).ok(
-      'Login dialog failed to appear'
-    )
```
danchr: Don't these changes effectively disable the test? It seems to me we should perhaps mock authentication during the test run to require certain dummy values.

mora/app.py
```diff
@@ -8,14 +8,16 @@
...
+import werkzeug
+from flask_session import Session
 
-from . import auth
 from . import cli
 from . import exceptions
 from . import service
+from . import settings
 from . import util
+from .auth import auth, sso
```
danchr: `auth.auth` feels kind of… iffy. Could we perhaps name the module something that's ever so slightly less repetitive?

mora/app.py
```diff
@@ -24,8 +26,19 @@
 app = flask.Flask(__name__, root_path=distdir, template_folder=templatedir)
 app.cli = cli.group
 
+# Session setup
+app.config.update({
+    'SESSION_TYPE': 'filesystem',
+    'SESSION_PERMANENT': False,
+    'SESSION_FILE_DIR': settings.SESSION_FILE_DIR
+})
```
danchr: This is the first instance of using actually Flask's configuration in MO, I believe. It would be nice if could move the code towards a more customary use of Flask, where we actually generated the `flask.Flask` object in a method, and used their nice configuration rather than our own JSON file.

mora/app.py
```diff
@@ -24,8 +26,19 @@
...
+Session(app)
```
danchr: This is an example of why I prefer module-level imports rather than individual fields. `Session` is a rather general name, also used by `requests`, for example, whereas `flask_session.Session()` is quite unambiguous.

mora/auth/auth.py
```diff
@@ -56,6 +57,20 @@ def __call__(self, r):
...
+@blueprint.route('/user', methods=['GET'])
+def get_user():
+    '''Get the currently logged in user
+
+    .. :quickref: Authentication; Get user
+
+    :return: The username of the user who is currently logged in.
+    '''
+
+    username = flask.session.get('username') or 'N/A'
```
danchr: Wouldn't it be better to return `None` or `null` when not logged in, and then have the frontend handle it?

mora/auth/auth.py
```diff
@@ -45,8 +46,8 @@ def __init__(self, assertion=None):
     def __call__(self, r):
         if self.assertion:
             assertion = self.assertion
-        elif flask.request:
-            assertion = flask.request.cookies.get(COOKIE_NAME)
+        elif flask.session and flask.session.get(COOKIE_NAME):
+            assertion = flask.session[COOKIE_NAME]
```
danchr: You can simplify this code to:

```python
        elif flask.session:
             assertion = flask.session.get(COOKIE_NAME)
```

There's no need to avoid dealing with a `None` value, only to use `None` instead…

Also, we should rename the variable `COOKIE_NAME` to something like `TOKEN_KEY` now that it no longer names an actual cookie 🙂

mora/auth/auth.py
```diff
@@ -89,25 +104,17 @@ def login():
...
 
     if assertion:
-        resp.set_cookie(
-            COOKIE_NAME, assertion,
-            secure=flask.request.is_secure,
-            httponly=True,
-        )
-    else:
-        resp.delete_cookie(COOKIE_NAME)
+        flask.session[COOKIE_NAME] = assertion
```
danchr: Is it possible to hit the conditional above? In general, it seems to me that we should either assign `None` to the session key or clear it whenever authentication fails.

mora/auth/auth.py
```diff
@@ -116,7 +123,19 @@ def logout():
...
-    response = flask.make_response()
-    response.delete_cookie(COOKIE_NAME)
-
-    return response
+    flask.session.clear()
+    return flask.make_response()
+
+
+@blueprint.route('/login', methods=['GET'])
+def get_login():
```
danchr: `@util.restrictargs()`?

mora/auth/tokens.py
```diff
@@ -14,9 +14,9 @@
 import flask
 import requests
 
-from . import exceptions
-from . import settings
-from . import util
+from mora import exceptions
```
danchr: `from ..`?

tests/test_auth.py
```diff
@@ -22,6 +22,11 @@
 )
 
 
+# FIXME: itsdangerous.py has EPOCH hardcoded to 2011/01/01 for whatever reason,
+#        which breaks when using dates earlier than that.
+#        Fix has been implemented, awaiting next release:
+#        https://github.com/pallets/itsdangerous/issues/46
+@patch('itsdangerous.EPOCH', 0)
```
danchr: Erm, does the test really rely on a date that early?

cmoesgaard: They did, `@freezegun.freeze_time('2001-01-01')`
However, I can't seem to reproduce the error now. Weird.
I'll remove it and see how it goes.

danchr: My point was rather that the date likely needn't be _that_ earlier — hopefully `20110102` would suffice?

e2e-tests/LoginTest.js
cmoesgaard: The login page no longer features a login modal, so I simply removed that part from the test.
As far as I can tell, we test as much now, as we did before.

mora/app.py
cmoesgaard: Absolutely, that would be nice. Let's do that at some point (tm)

mora/settings.py
```diff
@@ -33,6 +33,8 @@
 SSO_SAML_METADATA_URL = '/url/to/sso/metadata'
 SSO_SAML_USERNAME_ATTR = ''
 
+SAML_IDP_INSECURE = True
```
danchr: You're overwriting the [previous assignment](https://github.com/magenta-aps/mora/pull/134/commits/b37409d1faa9a3f24579476243c5d66cbeebeecf#diff-99faeed6f80f83309c4eb1d6f068d108R30) to this variable.

mora/auth/base.py
```diff
@@ -45,8 +46,8 @@ def __init__(self, assertion=None):
     def __call__(self, r):
         if self.assertion:
             assertion = self.assertion
-        elif flask.request:
-            assertion = flask.request.cookies.get(COOKIE_NAME)
+        elif flask.session and flask.session.get(TOKEN_KEY):
+            assertion = flask.session[TOKEN_KEY]
```
danchr: A suggestion:

```python
        elif flask.session:
             assertion = flask.session.get(TOKEN_KEY)
```

…assuming the session can be falsy, that is.

mora/auth/base.py
```diff
@@ -56,6 +57,20 @@ def __call__(self, r):
...
+@blueprint.route('/user', methods=['GET'])
+def get_user():
+    '''Get the currently logged in user
+
+    .. :quickref: Authentication; Get user
+
+    :return: The username of the user who is currently logged in.
+    '''
+
+    username = flask.session.get('username') or None
```
danchr:
```python
return flask.session.get('username')
```
?

mora/auth/base.py
```diff
@@ -89,25 +104,19 @@ def login():
...
-            secure=flask.request.is_secure,
-            httponly=True,
-        )
+        flask.session[TOKEN_KEY] = assertion
     else:
-        resp.delete_cookie(COOKIE_NAME)
+        flask.session.clear()
 
-    return resp
+    return flask.make_response()
```
danchr: I was about to complain about this, but actually, it's vaguely more clear than a mere:

```python
    return ''
```

src/components/MoNavbar.vue
```diff
@@ -70,7 +71,7 @@
         this.$store.dispatch(AUTH_LOGOUT, vm.user)
         .then(response => {
           vm.isLoading = false
-          vm.$router.push({name: 'Login'})
+          window.location.replace('/')
```
danchr: Perhaps we should do the login prompt in a popup? That's vaguely nicer, and quite customary.

cmoesgaard: I'm not exactly sure what you mean. Have Flask generate a login popup?

danchr: > I'm not exactly sure what you mean. Have Flask generate a login popup?

The usual workflow is that the application triggers a popup at the remote site, you do the login, and redirect back to a page in the application. Due to having the same origin, that page can access its opener, and inform it that a successful login occurred. The main justification is that this allows you to retain application state during the login.

PR #137, opened 2018-08-23
#23559 end times

danchr:
This adjusts end times so that the frontend renders them properly, i.e. inclusive rather than exclusive.

The incorrect behaviour:
screenshot-adbee5b2-1b32-410b-85fc-f7a67a7be75c

### DIFF
--- a/mora/util.py
+++ b/mora/util.py
@@ -118,6 +118,19 @@ def to_lora_time(s):
         return dt.isoformat()
 
 
+def adjust_end_date(dt):
+    if (
+        dt not in (POSITIVE_INFINITY, NEGATIVE_INFINITY) and
+        dt.time() >= datetime.time(23, 59)
+    ):
+        return datetime.datetime.combine(
+            dt.date() + datetime.timedelta(days=1),
+            datetime.time(tzinfo=dt.tzinfo),
+        )
+    else:
+        return dt
+
+
 def to_iso_time(s):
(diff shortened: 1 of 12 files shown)


mora/validator.py
```diff
@@ -53,7 +53,8 @@ def _is_date_range_valid(parent: typing.Union[dict, str],
             if startdate < start:
                 # start is too late!
                 return False
-        elif start != previous_end:
+        elif start not in (previous_end,
+                           previous_end + datetime.timedelta.resolution):
```
cmoesgaard: Why is this necessary?

mora/util.py
```diff
@@ -118,6 +118,19 @@ def to_lora_time(s):
         return dt.isoformat()
 
 
+def adjust_end_date(dt):
```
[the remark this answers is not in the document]
cmoesgaard: We should also make sure to note that similar related time logic exists in the UI code.

mora/util.py
```diff
@@ -118,6 +118,19 @@ def to_lora_time(s):
...
+    if (
+        dt not in (POSITIVE_INFINITY, NEGATIVE_INFINITY) and
+        dt.time() >= datetime.time(23, 59)
```
cmoesgaard: Remember to adjust to be as specific as possible.

danchr:
The general approach with ε and subtracting something small is just too unwieldy — simply assuming midnight and dealing in dates instead should be much more maintainable.

PR #138, opened 2018-08-23
facet: include manager address in output

danchr:


mora/service/facet.py
```diff
@@ -51,6 +51,7 @@
     keys.ROLE_TYPE: 'Rolletype',
     # keys.USER_TYPE: 'Brugertype',
     keys.LEAVE_TYPE: 'Orlovstype',
+    keys.MANAGER_ADDRESS: 'Lederadresse',
```
cmoesgaard: I'd rather we call this `MANAGER_ADDRESS_TYPE` so also signify that this is essentially an alternative list of address types, and can be used as such in the `address_type` 'slot' in the manager payloads.

Also, we should update the docs with this information.

PR #148, opened 2018-09-04
Feature/22381 bvn path

robertjensen:
An attempt at an implementation of the feature request described in Redmine 22381.

The implementation adds an extra row to the response from get_one_orgunit with is populated with the path to the unit. The UI is updated to show this value instead of bvn. Suitable updates has been made to the tests to allow them to accept the extra row in the response.


mora/service/orgunit.py
```diff
@@ -205,6 +205,16 @@ def get_one_orgunit(c, unitid, unit=None,
     elif details is UnitDetails.FULL:
         unittype = common.get_uuid(rels['enhedstype'][0], required=False)
 
+        if rels['overordnet'][0]['uuid']is not None:
+            parent = get_one_orgunit(c, rels['overordnet'][0]['uuid'],
+                                     details=UnitDetails.FULL)
+            if parent is not None:
+                path = parent['path'] + '/' + parent['name']
+            try:
```
danchr: Generally speaking, `NameError`, like `SyntaxError`, shouldn't be raised as part of normal execution of the code. Given that you only assign to `path` in the prior conditional, please fold this `try`/`except` into the conditional prior to it.

mora/service/orgunit.py
```diff
@@ -205,6 +205,16 @@ def get_one_orgunit(c, unitid, unit=None,
     elif details is UnitDetails.FULL:
         unittype = common.get_uuid(rels['enhedstype'][0], required=False)
 
+        if rels['overordnet'][0]['uuid']is not None:
+            parent = get_one_orgunit(c, rels['overordnet'][0]['uuid'],
```
danchr: Since we're obtaining the parent, we might as well include it in the output, since it might be useful for other stuff as well. So just assign this to `r['parent']` instead.

src/organisation/OrganisationDetail.vue
```diff
@@ -7,7 +7,7 @@
 
       <div class="row">
         <div class="col">
-          <p class="card-text">Enhedsnr.: {{orgUnit.user_key}}</p>
+          <p class="card-text">Enhedsnr.: {{orgUnit.path}}</p>
```
danchr: I think we should include this _in addition_ to the user key, rather than _instead of_ it for now. If this customer truly wants to hide the user key, we should do that in configuration. Other customers might consider the user-key/short name useful

tests/test_integration_org_unit.py
```diff
@@ -22,13 +22,14 @@ class Tests(util.LoRATestCase):
 
     def test_org_unit_temporality(self):
         self.load_sample_structures()
-
```
danchr: Erm, newlines are good — they make the code more readable. Any particular reason to kill this one?

tests/test_integration_org_unit.py
```diff
@@ -22,13 +22,14 @@ class Tests(util.LoRATestCase):
...
         self.assertRequestResponse(
             '/service/ou/04c78fc2-72d2-4d02-b55f-807af19eac48'
             '/details/org_unit?validity=past',
             [
                 {
                     "name": "Afdeling for Fremtidshistorik",
+                    'path': ('/Overordnet Enhed/Humanistisk fakultet/' +
```
danchr: In the original customer request, the path didn't start with a `/`, did it? Also, the path doesn't include the name of the unit itself — is this intentional?

tests/test_integration_service.py
```diff
@@ -506,6 +506,8 @@ def test_orgunit(self):
                     'uuid': '32547559-cfc1-4d97-94c6-70b192eff825',
                 },
                 'parent': None,
+                'path': '',
```
danchr: Any particular reason why root units don't have a path?

mora/service/orgunit.py
```diff
@@ -205,6 +205,16 @@ def get_one_orgunit(c, unitid, unit=None,
     elif details is UnitDetails.FULL:
         unittype = common.get_uuid(rels['enhedstype'][0], required=False)
 
+        if rels['overordnet'][0]['uuid']is not None:
```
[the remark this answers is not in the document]
robertjensen: Fixed. Quite surprised that this does not give a syntax error...

mora/service/orgunit.py
robertjensen: This I new to me. My understanding of Python is that exceptions are routinely used to mark the end of an iteration? In this case it indicates that the recursion has reached the top level. I can write it in another way if you want, but I think this is a quite common way to do it.

https://www.python.org/dev/peps/pep-0234/
https://jeffknupp.com/blog/2013/02/06/write-cleaner-python-use-exceptions/

src/organisation/OrganisationDetail.vue
robertjensen: As I understood Alex, the chosen layout should be used until somebody explicitly wants something else, and then it will need to go into configuration. Ballerup do not want the user_key to be displayed.

tests/test_integration_org_unit.py
robertjensen: I have no opinions on this and will put it back. I do not know why it was removed, most likely old habit because I am used to another convention for this.

tests/test_integration_org_unit.py
robertjensen: I actually originally returned the answer without the starting slash:
https://github.com/magenta-aps/mora/commit/9e7510b5bf849c05968d79291bdea67e7465da96#diff-40eda68fe139c8fe75670d0531e68967R215
But then it was requested that the user_key should still be returned by the rest-interface and then I forgot about it. I will get rid of it again (and update all tests....)

tests/test_integration_service.py
robertjensen: The convention I have followed is that the name of the ou itself should not be part of the path. The name of the ou is written immediately above, and thus it would look silly, to repeat it + it keeps the path slightly shorter and thus avoids line-breaks in more cases. In this convention, the root ou will have no path.

mora/service/orgunit.py
robertjensen: ...ahhh now I see it...ok in this case I can do without the exception :-)
But I still do not agree that exceptions should not be part of normal execution.

tests/test_integration_service.py
danchr: > The convention I have followed is that the name of the ou itself should not be part of the path. The name of the ou is written immediately above, and thus it would look silly, to repeat it + it keeps the path slightly shorter and thus avoids line-breaks in more cases. In this convention, the root ou will have no path.

Usually, the path to an item includes the name of that path; for example, a file system path _always_ includes the final item. I think they consider this a form of identification as well, in which case they almost certainly want to include the item itself.

mora/service/orgunit.py
danchr: Well, I think you misunderstood me; I did not mean to say that _all_ exceptions should be avoided, but rather than specifically `NameError` and `SyntaxError` 🙂 

`StopIteration` is an interesting case, by the way, and has some rather odd side effects — at least without a `__future__` import, see [PEP-479](](https://www.python.org/dev/peps/pep-0479/). (We should consider adding that `__future__` import…)

Speaking of exceptions, this is a common antipattern:

```python
try:
  x = mapping[42]
except KeyError:
  x = 1337
```

Instead, do this:

```python
  x = mapping.get(42, 1337)
```

…but that doesn't imply that `KeyError` is evil or anything like that — it's very useful for non-trivially handling a missing value.

mora/service/orgunit.py
```diff
@@ -205,14 +205,13 @@ def get_one_orgunit(c, unitid, unit=None,
...
-                path = parent['path'] + '/' + parent['name']
-            try:
-                r['path'] = path
-            except NameError:
+        if rels['overordnet'][0]['uuid'] is not None:
+            r['parent'] = get_one_orgunit(c, rels['overordnet'][0]['uuid'],
+                                          details=UnitDetails.FULL)
+            if r['parent'] is not None:
+                slash = '' if not r['parent']['path'] else '/'
+                r['path'] = r['parent']['path'] + slash + r['parent']['name']
```
danchr: If you include the current item in the path, you can simplify these two lines to simply be:

```python
 r['path'] = '/'.join((r['parent']['path'], r['name']))
```

mora/service/orgunit.py
```diff
@@ -237,7 +236,6 @@ def get_one_orgunit(c, unitid, unit=None,
         )
 
     r[keys.VALIDITY] = validity or common.get_effect_validity(validities[0])
-
```
danchr: I still like newlines 🙂

mora/service/orgunit.py
```diff
@@ -205,6 +205,16 @@ def get_one_orgunit(c, unitid, unit=None,
     elif details is UnitDetails.FULL:
         unittype = common.get_uuid(rels['enhedstype'][0], required=False)
 
+        if rels['overordnet'][0]['uuid'] is not None:
+            r['parent'] = get_one_orgunit(c, rels['overordnet'][0]['uuid'],
+                                          details=UnitDetails.FULL)
+
+            if r['parent'] is not None:
+                slash = '' if not r['parent']['path'] else '/'
+                r['path'] = r['parent']['path'] + slash + r['parent']['name']
```
danchr: It just so happens that this statement has a quadratic execution time — but it [probably](https://stackoverflow.com/questions/4435169/how-do-i-append-one-string-to-another-in-python#4435194) won't matter…

That being said, this code region is a bit hard to read — I think it'd be cleaner if you used explicit `if`s for all three cases.

NEWS.rst
```diff
@@ -5,15 +5,16 @@ New features
 ------------
 
 * #22849: Implemented support for signed SAML AuthN Requests.
-
+* #22381: Replace 'Enhedsnummer' with the a desription the OUs position in the tree
```
cmoesgaard: Bunch of typos.
`#22381: Replace 'Enhedsnummer' with the description the OUs position in the tree`

danchr: I'd suggest `with a description of the location of the organisational unit`

danchr:
I'd appreciate the final change, but overall it looks good!

PR #162, opened 2018-09-11
24067: Fixed so we are now able to edit root units

cmoesgaard:
https://redmine.magenta-aps.dk/issues/24067

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -7,6 +7,12 @@ New features
 * #22849: Implemented support for signed SAML AuthN Requests.
 * #22381: Replace 'Enhedsnummer' with a description of the location of the organisational unit
   
+
+Bug fixes
+---------
+
+* #24067: Fixed being able to edit root organisational units
+
 Version 0.9.0, 2018-09-07
 =========================
 
--- a/backend/mora/service/orgunit.py
+++ b/backend/mora/service/orgunit.py
@@ -613,7 +613,7 @@ def edit_org_unit(unitid):
     to contain the fields that need to change along with the validity dates.
 
     :<json string name: The name of the org unit
-    :<json string parent: The parent org unit
+    :<json string parent: The parent org unit or organisation
     :<json string org_unit_type: The type of org unit
     :<json object validity: The validities of the changes.
 
--- a/backend/mora/service/orgunit.py
+++ b/backend/mora/service/orgunit.py
@@ -162,6 +162,7 @@ class OrgUnit(common.AbstractRelationDetail):
         payload = common.ensure_bounds(new_from, new_to, bounds_fields,
                                        original, payload)
 
+        # TODO: Check if we're inside the validity range of the organisation
         if org_uuid != parent_uuid:
             validator.is_date_range_in_org_unit_range(parent_uuid, new_from,
                                                       new_to)
@@ -613,7 +614,7 @@ def edit_org_unit(unitid):
     to contain the fields that need to change along with the validity dates.
 
     :<json string name: The name of the org unit
-    :<json string parent: The parent org unit or organisation
+    :<json string parent: The parent org unit
     :<json string org_unit_type: The type of org unit
     :<json object validity: The validities of the changes.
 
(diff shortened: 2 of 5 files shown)


NEWS.rst
```diff
@@ -6,6 +6,11 @@ New features
 
 * #22849: Implemented support for signed SAML AuthN Requests.
 
+Bug fixes
+---------
+
+* #24067: Fixed being able to edit root organisational units
```
danchr: “Allowed editing root organisational units.”?

Strictly speaking, I wouldn't say this was a bug, but rather a missing feature.

backend/mora/service/orgunit.py
```diff
@@ -106,6 +106,10 @@ def edit(self, unitid, req):
             original, mapping.PARENT_FIELD.path)[-1]
         parent_uuid = util.get_uuid(parent)
 
+        org = util.get_obj_value(
```
danchr: Don't we have a `get_obj_uuid` (or something) for exactly this case?

backend/mora/service/orgunit.py
```diff
@@ -158,8 +162,9 @@ def edit(self, unitid, req):
         payload = common.ensure_bounds(new_from, new_to, bounds_fields,
                                        original, payload)
 
-        validator.is_date_range_in_org_unit_range(parent_uuid, new_from,
-                                                  new_to)
+        if org_uuid != parent_uuid:
+            validator.is_date_range_in_org_unit_range(parent_uuid, new_from,
+                                                      new_to)
```
danchr: We should probably validate that we're not exceeding the organisation's validity, shouldn't we?

backend/mora/service/orgunit.py
```diff
@@ -593,7 +598,7 @@ def edit_org_unit(unitid):
     to contain the fields that need to change along with the validity dates.
 
     :<json string name: The name of the org unit
-    :<json string parent: The parent org unit
+    :<json string parent: The parent org unit or organisation
```
danchr: This seems to indicate that we can move units to/from the root level. Should that be the case? Do we test it?

NEWS.rst
cmoesgaard: We never mentioned _not_ being able to edit root organisational units, and the error message wasn't particularly helpful - so I think this qualifies as a bug.

backend/mora/service/orgunit.py
cmoesgaard: We have a ton of `get_*` functions, but not one for this particular purpose, AFAIK.

backend/mora/service/orgunit.py
cmoesgaard: I think organisations are always just created to last forever, if we're suddenly outside the validity period of the organisation, we have bigger problems.

backend/mora/service/orgunit.py
cmoesgaard: Yes, I don't see why not. And no, but I'll add some.

backend/mora/service/orgunit.py
danchr: Well, that depends on the importer — for example, most municipalities have a from day one might use. A simple `TODO` might suffice?

backend/mora/service/orgunit.py
danchr: My main worry is that whatever we do, we:

1) Know we're doing it.
2) Have tests for it.
3) On some level, it makes sense.

I'm willing to compromise on the last item as long as the two previous are covered 😁  But yes, moving from the root level makes sense, but is somewhat hard to test as it presumes a different hierarchy. Given that the easiest way to test it is moving a unit back and forth, we might as well do both IMO.

cmoesgaard: So you're suggensting we remove the existing validiation rule preventing us from moving from the root level?

danchr: I think we should ask @alexthirifays — they change seems fairly simple, but I can't quite settle on whether the users would want it. They probably would.

danchr: And the conclusion from asking him is that it seems a rather risky operation, so out of an abundance of caution we should disallow it, for now. In other words: Re-parenting a unit to or from the root is forbidden.

cmoesgaard: Alright. I'll add an additional validation to ensure we aren't able to move _to_ the root level as well.

danchr:
🏇

PR #163, opened 2018-09-11
#23559: switch to using dates in REST API

danchr:
We now use ISO 8601 dates in the API, rather than dates with time.
When reading from LoRA, we simply disregard any date. When validating
requests, anything other than midnight CE[S]T causes a hard error.

This works rather well in practice, although it _may_ lead to issues
should other clients of LoRA write an address or IT system with a
validity changeover not at midnight. In this case, we should be able
to read the data, but not alter it.

https://redmine.magenta-aps.dk/issues/23559

### DIFF
--- a/backend/mora/service/orgunit.py
+++ b/backend/mora/service/orgunit.py
@@ -842,7 +842,7 @@ def terminate_org_unit(unitid):
     c = lora.Connector(effective_date=util.to_iso_date(date))
 
     validator.is_date_range_in_org_unit_range(
-        unitid, date - util.ONE_DAY, date,
+        unitid, date - util.MINIMAL_INTERVAL, date,
     )
 
     children = c.organisationenhed.paged_get(
--- a/backend/mora/service/orgunit.py
+++ b/backend/mora/service/orgunit.py
@@ -839,8 +839,7 @@ def terminate_org_unit(unitid):
     """
     date = util.get_valid_from(flask.request.get_json(), is_end=True)
 
-    c = lora.Connector(virkningfra=util.to_iso_date(date),
-                       virkningtil='infinity')
+    c = lora.Connector(effective_date=util.to_iso_date(date))
 
     validator.is_date_range_in_org_unit_range(
         unitid, date - util.ONE_DAY, date,
--- a/backend/mora/service/orgunit.py
+++ b/backend/mora/service/orgunit.py
@@ -825,7 +825,7 @@ def terminate_org_unit(unitid):
 
       {
         "validity": {
-          "from": "2015-12-31"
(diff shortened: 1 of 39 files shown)


backend/mora/service/orgunit.py
```diff
@@ -837,13 +837,12 @@ def terminate_org_unit(unitid):
       }
 
     """
-    date = util.get_valid_from(flask.request.get_json())
+    date = util.get_valid_from(flask.request.get_json(), is_end=True)
 
-    c = lora.Connector(virkningfra=util.to_iso_time(date),
-                       virkningtil='infinity')
+    c = lora.Connector(effective_date=util.to_iso_date(date))
```
cmoesgaard: Actually, we could just do `date + util.ONE_DAY` here, and do away with the `is_end` parameter above, altogether - and similarly for terminating employees.

backend/mora/util.py
```diff
@@ -157,28 +153,40 @@ def to_iso_time(s):
     )
 
 
+def to_iso_date(s, is_end=False):
```
cmoesgaard: Explain what the `is_end` parameter does.

backend/mora/util.py
```diff
@@ -616,7 +635,15 @@ def get_valid_to(obj, fallback=None) -> datetime.datetime:
...
-            return from_iso_time(valid_to)
+            dt = from_iso_time(valid_to)
+
+            if dt.time() != datetime.time.min:
+                raise exceptions.HTTPException(
+                    exceptions.ErrorCodes.E_INVALID_INPUT,
+                    '{!r} is not at midnight!'.format(dt.isoformat()),
+                )
+
+            return dt + ONE_DAY
```
cmoesgaard: Maybe add a comment explaining why we're doing this?

backend/mora/service/orgunit.py
cmoesgaard: My point was, having a parameter called `is_end` for a function used to get a _from_ date, is slightly confusing in my opinion - it's not clear to me how to actually interpret this semantically, other than the parameter just being "return the date as usual, but with one day added"
The only places we actually call `get_valid_from` with this parameter is in these termination methods, where we still do a bunch of date arithmetic anyway.
So, to me this is an opportunity to get rid of a parameter to a function, keeping it simpler.

danchr: Well, as an alternative we could make add a parameter to the function for specifying the key — which is `from` rather than the customary `to` in this case. Would that make more sense?

danchr: For future reference: we changed the API instead.

PR #164, opened 2018-09-14
Feature/22849 saml sso

cmoesgaard:
https://redmine.magenta-aps.dk/issues/23485


backend/mora/auth/sso.py
```diff
@@ -5,38 +5,176 @@
...
+
+@blueprint.route('/acs/', methods=['POST'])
+@init_saml_auth
+def acs(auth):
+    auth.process_response()
+    errors = auth.get_errors()
+
+    if len(errors) == 0:
+        # TODO: Kill this when LoRa has better auth
+        res = OneLogin_Saml2_Utils.b64decode(
```
danchr: Use the [included module](https://docs.python.org/3/library/base64) instead?

backend/mora/auth/sso.py
```diff
@@ -5,38 +5,176 @@
...
-    """Custom assertion consumer service for use with flask_saml"""
-    flask.session[base.TOKEN_KEY] = tokens.pack(auth.xmlstr)
-    flask.session['username'] = attributes[settings.SSO_SAML_USERNAME_ATTR][0]
+@blueprint.route('/slo/')
+@init_saml_auth
+def slo(auth):
+    name_id = None
+    session_index = None
+    if 'samlNameId' in flask.session:
+        name_id = flask.session['samlNameId']
```
danchr:
```python
name_id = flask.session.get('samlNameId', None)
```

backend/mora/auth/sso.py
```diff
@@ -5,38 +5,176 @@
...
+    if 'samlSessionIndex' in flask.session:
+        session_index = flask.session['samlSessionIndex']
```
danchr: Likewise.

backend/mora/auth/sso.py
```diff
@@ -5,38 +5,176 @@
...
-        'SAML_KEY_FILE': settings.SAML_KEY_FILE,
-        'SAML_CERT_FILE': settings.SAML_CERT_FILE,
-        'SAML_AUTHN_REQUESTS_SIGNED': settings.SAML_AUTHN_REQUESTS_SIGNED,
-    })
+@blueprint.route('/sls/')
+@init_saml_auth
+def sls(auth):
+    url = auth.process_slo(delete_session_cb=delete_session_callback)
+    errors = auth.get_errors()
+    if len(errors) == 0:
```
danchr:
```python
if errors:
```

backend/mora/auth/sso.py
```diff
@@ -5,38 +5,176 @@
...
+        if url is not None:
+            return flask.redirect(url)
+        return flask.redirect('/')
+    else:
+        return flask.jsonify(errors)
```
danchr: Is HTTP 200 OK the appropriate response here? Perhaps 401?

backend/mora/settings.py
```diff
@@ -19,7 +19,8 @@
 LORA_URL = 'http://mox.lxc/'
 CA_BUNDLE = None
 
-AUTH = 'token'  # 'sso' or 'token'
+# Turn on SSO auth module
```
danchr: Erm, the comment here isn't especially helpful, and seems to disagree with the code?

backend/requirements.txt
```diff
@@ -13,6 +13,6 @@ pyexcel-ods>=0.5.2
 gevent~=1.3.5
 python-dateutil>=2.7.0
 setuptools>=40.0.0
--e git+https://bitbucket.org/magenta-aps/flask_saml@development#egg=flask_saml
+-e git+https://github.com/magenta-aps/python3-saml@attribute_fix#egg=python3-saml
```
danchr: What modifications do we have for `python3-saml`?

backend/mora/auth/sso.py
```diff
@@ -5,38 +5,176 @@
...
+
+
+@blueprint.route('/metadata/')
+@init_saml_auth
+def metadata(auth):
+    settings = auth.get_settings()
+    sp_metadata = settings.get_sp_metadata()
+    errors = settings.validate_metadata(sp_metadata)
+
+    if len(errors) == 0:
```
danchr: As earlier, and I'd consider flipping the condition both locations — that is, have the check for errors be explicit:
```python
if errors:
    return flask.jsonify(errors), 401

resp = …
```

backend/mora/auth/sso.py
```diff
@@ -5,38 +5,176 @@
...
+        'https': 'on' if flask.request.scheme == 'https' else 'off',
+        'http_host': flask.request.host,
+        'server_port': url_data.port,
+        'script_name': flask.request.path,
+        'get_data': flask.request.args.copy(),
+        'post_data': flask.request.form.copy()
+    }
+
+
+def init_saml_auth(func):
```
danchr: What does this do? Also, is `init` a good prefix?

backend/mora/auth/sso.py
```diff
@@ -5,38 +5,176 @@
...
+
+        s['sp'].update({
+            "x509cert": cert,
+            "privateKey": key
+        })
+
+    return s
+
+
+def prepare_flask_request():
```
danchr: `get_saml_request`?

backend/mora/auth/sso.py
```diff
@@ -5,38 +5,176 @@
...
+    # If server is behind proxys or balancers use the HTTP_X_FORWARDED fields
+    url_data = parse.urlparse(flask.request.url)
+    return {
+        'https': 'on' if flask.request.scheme == 'https' else 'off',
+        'http_host': flask.request.host,
```
danchr: Please note that this includes “the port if available”.

backend/requirements.txt
cmoesgaard: https://github.com/magenta-aps/python3-saml/commit/5277d78cf37b56979b0c2778b3e5589122f5e922

danchr: Reading the relevant bug, it seems to have security repercussions — what triggers this?

backend/mora/auth/sso.py
cmoesgaard: It creates and initializes a SAML client and passes it along to whatever function this is annotated by.
``init_saml_auth`` is probably not the best name. I'll think of something better.

backend/mora/auth/sso.py
cmoesgaard: :+1:

frontend/config/index.js
```diff
@@ -14,6 +14,10 @@ module.exports = {
       "/service": {
         target: process.env.BASE_URL || "http://localhost:5000/",
         changeOrigin: true
+      },
+      "/saml": {
```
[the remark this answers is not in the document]
cmoesgaard: I think having the Saml SSO endpoints at `/saml` is descriptive enough.

backend/mora/auth/sso.py
```diff
@@ -5,38 +5,272 @@
...
+        'server_port': url_data.port,
+        'script_name': flask.request.path,
+        'get_data': flask.request.args.copy(),
+        'post_data': flask.request.form.copy()
+    }
+
+
+def _prepare_saml_auth(func):
+    """Create and initialize OneLogin SAML2 Auth client"""
```
danchr: “Decorator to…” — otherwise one might reasonably assume that it was a normal function, or perhaps even a contextmanager. Speaking of which… [`contextlib.contextmanager`](https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager) is pretty neat, and ought to work as a decorator as well 🙂

backend/tests/test_sso.py
```diff
@@ -0,0 +1,206 @@
...
+            'SAMLResponse': self.get_slo_response(),
+        }
+
+        r = self.client.get('/saml/sls/', query_string=urlencode(data))
+
+        expected = {
+            'description': "['ERROR 2']",
+            'error': True,
+            'error_key': 'E_UNKNOWN',
+            'status': 500
```
danchr: Unfortunately, we cannot assume that a SAML error is, in fact, an error condition. For example, the cause of the error might simply be an incorrect login — see `tests/mocking/auth/adfs-failed-login.xml`, for example.

I think we should return 401 or 403 — or whatever we did previously — when the login fails.

backend/tests/test_sso.py
```diff
@@ -0,0 +1,206 @@
...
+                         query.get('RelayState')[0])
+
+    @patch('onelogin.saml2.response.OneLogin_Saml2_Response.is_valid',
+           lambda *x, **y: True)
+    def test_acs_redirects_correctly_with_relaystate(self):
+        data = {
+            'SAMLResponse': self.get_sso_response(),
+            'RelayState': 'http://redirect.me/to/here'
+        }
+        r = self.client.post('/saml/acs/', data=data)
```
danchr: AFAICT this test doesn't freeze the time. We should do so, and verify that the authentication fails when time is frozen at when the token either isn't valid yet or expired.

cmoesgaard: Good point, I'll do that.

backend/tests/test_sso.py
cmoesgaard: SSO works differently than what we used before, so if a user fails to authenticate they'll be stopped by the IdP, and thus we'll never receieve a failed login response such as `tests/mocking/auth/adfs-failed-login.xml`

backend/tests/test_sso.py
danchr: The general scenarios I can think of are:

1) Before the assertion is valid.
2) After the assertion expires.
3) When the assertion is signed by another entity/the signature is invalid.
4) When the assertion isn't issued to us.
5) Our behaviour on errors.

backend/tests/test_sso.py
danchr: Ah, that's good news — can you think of anything else that might trigger this? Did you see it during manual testing?

backend/mora/auth/sso.py
```diff
@@ -5,38 +5,273 @@
...
-from . import tokens
-from .. import settings
 
 basedir = os.path.dirname(__file__)
 
+blueprint = flask.Blueprint('sso', __name__, static_url_path='',
+                            url_prefix='/saml')
+
+
+class AuthException(Exception):
```
danchr: AFAICT you only instantiate this to raise it — shouldn't we use it, or check it, or something?

docs/conf.py
```diff
@@ -72,13 +56,35 @@
...
     'sphinxcontrib.autohttp.flaskqref',
+    'sphinxcontrib.apidoc',
 ]
 
 autodoc_default_flags = [
     # 'members',
     # 'undoc-members',
 ]
 
+MOCK_MODULES = [
```
danchr: Can't we use the built-in mocker now?

cmoesgaard: The built-in mocker behaves differently somehow, in a way I can't quite figure out.

PR #165, opened 2018-09-17
Feature/24053 vuedoc

n-singh:


danchr:
> > Shouldn't we include instructions on how to build the Vue docs? They're not built as part of the rest of the documentation in the makefile as far as I can see.
> 
> We should — it's worth noting that there's a test that builds them, and verifies that they're up-to-date.

I've reported issue [#24400](https://redmine.magenta-aps.dk/issues/24400) so that we keep track of this.

PR #170, opened 2018-09-24
Feature/24131 simple conf

robertjensen:
https://redmine.magenta-aps.dk/issues/24131

Did not write new tests, existing tests cover the fact the the module does not break functionality and new functionality is only visible in the UI, and thus hard to test. Existing tests are updated to accept the new syntax, though.

Documentation is currently in a separate file, awaiting merge of Christoffer's branch with updated docs.


backend/mora/settings.py
```diff
@@ -44,4 +44,13 @@
 
 PROD_MODE = False
 
+try:
+    with open('backend/mora/user_settings.json') as f:
```
cmoesgaard: The path for `user_settings` should optionally be configurable, e.g. with an environment variable.
This is in part a wish from the ops people, so the config file can exist separately from the code, which makes deployment and maintenance slightly easier.

(Ideally, we should do this for the location of the main settings file as well, but I think that's out of scope for this branch)

backend/mora/user_settings.json
```diff
@@ -1,4 +1,12 @@
-{'orgunit':
-   {'show_family_line': True,
-    'show_bvn': True}
+{
```
cmoesgaard: This file is described in the docs, and we have some sensible defaults, so I don't think we should include it.

configuration.rst
```diff
@@ -0,0 +1,42 @@
+Configuration module
```
cmoesgaard: There are a few typos
indcates, appply, examplel

docs/dev/configuration.rst
```diff
@@ -0,0 +1,43 @@
...
+It is possible to perform simple configuration of the MO frontend using the
+configuation file found in ``backend/mora/user_settings.json`` or in the file
+indicated in the env-variable USER_SETTINGS.
+
+This file could look something like this:
+
+    .. sourcecode:: json
+
+    {
+	"orgunit": {
```
danchr: That unindent completely breaks rendering — you can inspect the generated output using `make -C docs` with your venv activated.

frontend/src/organisation/OrganisationDetail.vue
```diff
@@ -2,12 +2,11 @@
   <div class="card">
     <div class="card-body">
       <h4 class="card-title">
-        <icon name="users" /> {{orgUnit.name}}
+        <icon name="users" /> {{orgUnitInfo.name}}
```
danchr: This renaming seems a bit arbitrary — from the perspective of the frontend, that object is indeed the unit.

backend/tests/test_integration_org_unit.py
```diff
@@ -33,6 +33,9 @@ def test_org_unit_temporality(self):
                     'location': ('Overordnet Enhed/Humanistisk fakultet/' +
                                  'Historisk Institut'),
                     "user_key": "frem",
+                    'user_settings': {'orgunit': {'show_bvn': False,
+                                                  'show_location': True,
+                                                  'show_roles': True}},
```
danchr: I'm not fond of including this with each and every request — it seems a bit too specific to the UI. Considering that we don't expect the configuration to change, we should consider just exposing it, and let the frontend determine whether to show anything. As it happens, you recently made the change to give it the entire tree, right?

docs/dev/configuration.rst
```diff
@@ -0,0 +1,43 @@
+Configuration module
+=====================
+
+Front-end configuration
+------------
+
+It is possible to perform simple configuration of the MO frontend using the
+configuation file found in ``backend/mora/user_settings.json`` or in the file
```
danchr: If we retain this, we should add, and mention, a `user_settings-example.json` just as we have `mora-example.json`.

frontend/src/organisation/OrganisationDetailTabs.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div>
+  <div v-if="settings">
```
danchr: Eh, why should we hide all tabs if we have no settings? That seems rather unusual…

PR #180, opened 2018-10-02
First run at decoupling validation flow for engagements

cmoesgaard:

### DIFF
--- a/backend/mora/service/detail_writing.py
+++ b/backend/mora/service/detail_writing.py
@@ -47,6 +47,17 @@ CREATION_HANDLERS = {
     'address': address.create_address
 }
 
+CREATE_VALIDATION_HANDLERS = {
+    'engagement': engagement.validate_create_engagement,
+    # 'association': association.edit_association,
+    # 'role': role.edit_role,
+    # 'leave': leave.edit_leave,
+    # 'manager': manager.edit_manager,
+    # 'it': itsystem.edit_itsystem,
+    # 'address': address.edit_address,
+    # 'org_unit': orgunit.edit_orgunit,
+}
+
 EDIT_HANDLERS = {
     'engagement': engagement.edit_engagement,
     'association': association.edit_association,
@@ -56,6 +67,17 @@ EDIT_HANDLERS = {
     'it': itsystem.edit_itsystem,
     'address': address.edit_address,
     'org_unit': orgunit.edit_orgunit,
+}
+
+EDIT_VALIDATION_HANDLERS = {
+    'engagement': engagement.validate_edit_engagement,
+    # 'association': association.edit_association,
+    # 'role': role.edit_role,
+    # 'leave': leave.edit_leave,
+    # 'manager': manager.edit_manager,
+    # 'it': itsystem.edit_itsystem,
+    # 'address': address.edit_address,
+    # 'org_unit': orgunit.edit_orgunit,
 }
 
 
@@ -357,10 +379,14 @@ def create():
 
     """
 
+    reqs = flask.request.get_json()
+
+    validated = process_requests(CREATE_VALIDATION_HANDLERS, reqs)
+
     return (
         flask.jsonify(
             process_requests(CREATION_HANDLERS,
-                             flask.request.get_json()),
+                             validated),
         ),
         201,
     )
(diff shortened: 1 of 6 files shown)


backend/mora/service/detail_writing.py
```diff
@@ -357,10 +379,14 @@ def create():
...
 
+    reqs = flask.request.get_json()
+
+    validated = process_requests(CREATE_VALIDATION_HANDLERS, reqs)
+
     return (
         flask.jsonify(
             process_requests(CREATION_HANDLERS,
-                             flask.request.get_json()),
+                             validated),
```
danchr: This feels somewhat weird — `validated` reads like a `bool`, but probably represents a preprocessed state or something similar?

backend/mora/service/employee.py
```diff
@@ -17,17 +17,17 @@
...
+import copy
 import uuid
 
 import flask
 
-from .. import mapping
-from .. import exceptions
-from . import address
-from .. import common
+from mora.service import detail_writing as writing
```
danchr: `from .` 🙂

backend/mora/service/employee.py
```diff
@@ -342,6 +342,8 @@ def create_employee():
     org_uuid = util.get_mapping_uuid(req, mapping.ORG, required=True)
     cpr = util.checked_get(req, mapping.CPR_NO, "", required=False)
     userid = util.get_uuid(req, required=False)
+    if not userid:
```
danchr: Perhaps we should add a `default` argument to `get_uuid()`?

backend/mora/service/employee.py
```diff
@@ -378,6 +380,31 @@ def create_employee():
...
     userid = c.bruger.create(user, uuid=userid)
 
-    return flask.jsonify(userid)
+    creation_uuids = writing.process_requests(writing.CREATION_HANDLERS,
+                                              validated_details)
+
+    return flask.jsonify([userid] + creation_uuids)
+
+
+def _decorate_create_payloads(details, employee_uuid, valid_from, valid_to):
```
danchr: `decorate` isn't quite the appropriate word — too similar to `decorator`, I'd say. Perhaps `_inject_persons`?

backend/mora/service/employee.py
```diff
@@ -378,6 +380,31 @@ def create_employee():
...
+    decorated = copy.deepcopy(details)
+    for detail in decorated:
+        detail['person'] = {
+            mapping.UUID: employee_uuid,
+            mapping.VALID_FROM: valid_from,
+            mapping.VALID_TO: valid_to,
+            'future': True
```
danchr: I know I suggested `future` (on a whim) but it doesn't feel right. `allow_nonexistent`? `created`?

backend/mora/service/engagement.py
```diff
@@ -48,8 +47,7 @@ def create_engagement(req):
                                                  mapping.ENGAGEMENT_TYPE,
                                                  required=True)
 
-    bvn = "{} {} {}".format(employee_uuid, org_unit_uuid,
-                            mapping.ENGAGEMENT_KEY)
+    bvn = str(uuid.uuid4())
```
danchr: We really ought to come up with something better here…

backend/mora/service/engagement.py
```diff
@@ -63,10 +61,21 @@ def create_engagement(req):
         opgaver=[{'uuid': job_function_uuid}] if job_function_uuid else []
     )
 
-    return c.organisationfunktion.create(engagement)
+    return {
```
danchr: In order to generalise this to addresses, you should probably return the path and/or scope as well.

backend/mora/validator.py
```diff
@@ -150,27 +150,47 @@ def is_distinct_responsibility(
...
+                exceptions.ErrorCodes.V_DATE_OUTSIDE_EMPL_RANGE,
+                employee_uuid=employee_uuid,
+                **_get_active_validity(employee),
+            )
 
-    if not _is_date_range_valid(employee, valid_from, valid_to, scope,
-                                gyldighed_key):
+
+def is_contained_in_employee_range(empl_from, empl_to, valid_from, valid_to):
+    if valid_from < empl_from or empl_to < valid_to:
```
danchr: This check is probably correct, but I've gotten checks like this subtly wrong, so we should make sure to unit test it thoroughly.

backend/mora/service/detail_writing.py
cmoesgaard: Yeah, it represents a set of payloads that have been through validation etc. and are ready to be sent to LoRa. I'll think of a better name.

PR #181, opened 2018-10-02
Feature/24135 sorted columns

n-singh:


danchr:
We already have some code in the backend that sorts the results — it just happens to sort by date, primarily. Perhaps we should adjust that logic instead?

PR #183, opened 2018-10-04
Feature/23976 separate validation

cmoesgaard:
https://redmine.magenta-aps.dk/issues/23976


backend/mora/service/address.py
```diff
@@ -415,11 +495,13 @@ def convert(effect):
                 addr[mapping.VALIDITY] = util.get_effect_validity(addrrel)
 
                 if self.scope.path == 'organisation/bruger':
+                    from . import employee
```
danchr: Erm, local imports are both a) slow and b) ugly. I'd rather see this in another module than `common` than a local import 😕

backend/mora/service/address.py
```diff
@@ -415,11 +495,13 @@ def convert(effect):
...
                     addr[mapping.PERSON] = employee.get_one_employee(
                         c, id, effect,
                     )
 
                 else:
+                    from . import orgunit
```
danchr: See earlier comment.

backend/mora/service/address.py
```diff
@@ -375,6 +373,88 @@ def get_one_address(c, addrrel, class_cache=None):
...
+
+    def edit(self, req: dict):
+        old_entry = util.checked_get(self.request, 'original', {},
+                                     required=True)
+        new_entry = util.checked_get(self.request, 'data', {}, required=True)
+
+        self.uuid, self.obj_type = get_id_and_type(old_entry)
+
+        get_relation_for(old_entry)
+        get_relation_for(new_entry, old_entry)
```
danchr: I'm not entirely fond of this approach; AFAICT you do everything once, to validate it, and then once more, to issue the request. That seems a bit wasteful and duplicative, to me.

backend/mora/common.py
```diff
@@ -50,6 +51,77 @@ def get(self, objid):
...
+        :param request: A dict containing a request
+        :param request_type: A RequestType, either CREATE or EDIT
+        """
+        self.request_type = request_type
+        self.request = request
+        self.payload = None
+        self.uuid = None
+
+        if request_type == RequestType.CREATE:
+            self.create(request)
```
danchr: Wouldn't it be better if the caller invoked these methods? I.e.:

```python
x = WhateverRequest(…).create(req)
```

backend/mora/common.py
```diff
@@ -50,6 +51,77 @@ def get(self, objid):
...
+        Submit the request to LoRa.
+        :return: A string containing the result from submitting the request
+        to LoRa, typically a UUID.
+        """
+        pass
+
+
+class OrgFunkRequest(Request):
+    @abc.abstractmethod
+    def create(self, request: dict):
```
danchr: The naming is a bit off here; `create` implies that the method creates something when in fact it does not. `prepare_creation`?

backend/mora/common.py
```diff
@@ -50,6 +51,77 @@ def get(self, objid):
...
+        pass
+
+    @abc.abstractmethod
+    def edit(self, request: dict):
```
danchr: `prepare_edit`?

backend/mora/common.py
```diff
@@ -50,6 +51,77 @@ def get(self, objid):
...
+        pass
+
+    def submit_request(self) -> str:
```
danchr: `submit`?

backend/mora/common.py
```diff
@@ -50,6 +51,77 @@ def get(self, objid):
         pass
 
 
+class RequestType(enum.Enum):
+    CREATE = 0
+    EDIT = 1
+
+
+class Request(abc.ABC):
```
danchr: To avoid confusing these for `requests.Request`, I'd suggest calling them `Handler` or something of that ilk.

backend/mora/common.py
danchr: (Also, you're not calling super().__init__() in your method…)

backend/mora/service/address.py
cmoesgaard: Yeah, I see your point. I'll find a suitable way of saving the intermediate result, as I do with the other handlers.

backend/mora/service/address.py
cmoesgaard: It was purely to resolve an iffy circular dependency. I'll try to move stuff around to avoid this.

backend/mora/common.py
cmoesgaard: We could let the caller invoke the `create` methods, but I don't see how it would be better. Just different.
And I'll add a call to `__init__()`.

frontend/src/api/Employee.js
```diff
@@ -70,11 +70,15 @@ export default {
   new (employee) {
     return Service.post('/e/create', employee)
       .then(response => {
+        let EmployeeUuid = response.data
```
danchr: They've been using `snake_case` until now.

backend/mora/common.py
```diff
@@ -50,6 +51,112 @@ def get(self, objid):
...
+class RequestType(enum.Enum):
+    CREATE = 0
+    EDIT = 1
+
+
+# Handlers is populated by each individual active RequestHandler
+HANDLERS = {}
+
+
+def register_request_handler(name):
```
danchr: `register` is redundant — actually, how about `handler_for()`?

backend/mora/common.py
```diff
@@ -56,6 +56,18 @@ class RequestType(enum.Enum):
     EDIT = 1
 
 
+# Handlers is populated by each individual active RequestHandler
```
danchr: Totally picking a nit, but “Handlers is” sounds really weird. “The "Handlers" mapping is”?

backend/mora/common.py
cmoesgaard: Well, I wanted it to be obvious that adding the decorator would actively register the decorated handler with the system.

danchr: It's just that this chunk of code feels somewhat repetitive:

```python
@common.register_request_handler('address')
 class AddressRequestHandler(common.RequestHandler):
```

Anyway, it's definitely a minor issue, so let's not hold up this change on it. Plus, you wrote the code, after all!

PR #190, opened 2018-10-08
#24150 manager termination

danchr:
https://redmine.magenta-aps.dk/issues/24150


backend/mora/common.py
```diff
@@ -52,18 +53,22 @@ def get(self, objid):
...
 
 
 # The handler mapping is populated by each individual active RequestHandler
 HANDLERS = {}
+HANDLERS_BY_KEY = {}
+FUNCTION_KEYS = {}
 
 
-def register_request_handler(name):
+def register_request_handler(name, key=None):
```
cmoesgaard: It feels wrong to add an additional parameter here, that only makes sense for org funk handlers. We have a class that every org funk handler inherits from, so maybe the logic regarding function keys could be moved there instead?

backend/mora/common.py
```diff
@@ -135,6 +178,14 @@ def submit(self) -> str:
             return c.organisationfunktion.update(self.payload, self.uuid)
 
 
+def handler_for(obj: dict):
```
cmoesgaard: Similarly here, this only concerns itself with org funks, despite the function being called `handler_for`. Rename it, or move the logic to the abstract org funk class somehow.

backend/mora/common.py
```diff
@@ -52,26 +53,48 @@ def get(self, objid):
...
-    return wrapper
 
 
-class RequestHandler(abc.ABC):
+class RequestHandler(metaclass=_RequestHandlerMeta):
 
     __slots__ = 'request', 'request_type', 'payload', 'uuid'
 
+    role_type = None
+    function_key = None
```
cmoesgaard: I still don't think it makes sense to have this here. Function key _only_ makes sense for org funk handlers, so the other handler types shouldn't know or care about it.

backend/mora/common.py
```diff
@@ -52,26 +53,48 @@ def get(self, objid):
...
+HANDLERS_BY_ROLE_TYPE = {}
+HANDLERS_BY_FUNCTION_KEY = {}
+FUNCTION_KEYS = {}
 
 
-def register_request_handler(name):
-    """Decorator to register a writing request handler"""
-    def wrapper(cls):
-        HANDLERS[name] = cls
+class _RequestHandlerMeta(abc.ABCMeta):
```
cmoesgaard: Using metaclasses seems unnecessarily complicated compared to using decorators as before, especially because now this code is run not only for the specific handler classes, but also for every abstract handler class as well.

backend/mora/common.py
```diff
@@ -22,6 +22,7 @@
 import datetime
 import enum
 import functools
+import operator
```
cmoesgaard: Not used

PR #197, opened 2018-10-12
#24069 handle missing addresses

danchr:
https://redmine.magenta-aps.dk/issues/24069

- [ ] ~Have you performed a smoke test of the application?~


backend/mora/service/address.py
```diff
@@ -312,14 +314,44 @@ def get_one_address(c, addrrel, class_cache=None):
...
+                mapping.NAME: DEFAULT_ERROR,
+                mapping.UUID: addrrel['uuid'],
+                mapping.ERROR: str(exc),
+            }
 
-        r.raise_for_status()
+        if not r.ok:
+            error = r.json()
+
+            flask.current_app.logger.warn(
```
cmoesgaard: I've often thought that we should use the logger more, so this is good.

backend/mora/service/address.py
```diff
@@ -312,14 +314,44 @@ def get_one_address(c, addrrel, class_cache=None):
...
+                    'noformat': '1',
+                },
+            )
+        except Exception as exc:
+            # the exception above is overly broad for a) safety and b)
+            # testing -- specifically, the exception raised by
+            # requests_mock does not descend from RequestException :(
+            util.log_exception('failed to get address ' + addrrel['uuid'])
+
+            return {
```
cmoesgaard: We build this object four different places now. Maybe it should be extracted to a function, to ensure consistency of the objects being created?

backend/mora/service/address.py
```diff
@@ -312,6 +312,15 @@ def get_one_address(c, addrrel, class_cache=None):
     scope = util.checked_get(addrclass, 'scope', 'DAR')
 
     if scope == 'DAR':
+        def make_error_object(errordesc):
```
cmoesgaard: Why not replace _all_ of the places we create these address objects? Specifically lines 355 and 392. They all need to have the exact same fields (except for the addition of `ERROR` which is optional), so why leave it up to chance by constructing the dictionaries by each time?

PR #202, opened 2018-10-24
unbreak Jenkins by removing our explicit requirement for pycodestyle

danchr:
This was a work-around for an earlier version of flake8 being broken
with a newer pycodestyle.


danchr:
> I'd rather we just fix the errors later, than add ignores now.

I don't quite follow… The errors currently break the build; what should we do other than ignoring them?

cmoesgaard:
Stay on the version of `flake8` known to work, until we fix the ones preventing us from using the newest version.
I'd rather we do that, than introducing code for parsing configs and handling flake8 ignores, that we won't need when the issues are fixed. Less code is better than more code.

PR #204, opened 2018-10-25
edit behaviour for employee object

jgj-magenta-aps:
- [X] Have you updated the documentation?
- [X] Have you performed a smoke test of the application?
- [X] Have you written tests for your changes?
- [X] Have you updated the release notes?


backend/tests/test_integration_employee.py
```diff
@@ -458,3 +458,180 @@ def test_cpr_lookup_raises_on_wrong_length(self):
             },
             status_code=400,
         )
+
+    def test_edit_employee_overwrite(self):
```
cmoesgaard: This test should assert that `cpr_no` has been changed correctly as well.

backend/tests/test_integration_employee.py
```diff
@@ -458,3 +458,180 @@ def test_cpr_lookup_raises_on_wrong_length(self):
...
+        self.assertEqual(
+            expected_brugeregenskaber,
+            actual['attributter']['brugeregenskaber']
+        )
+        self.assertEqual(
+            expected_brugergyldighed,
+            actual['tilstande']['brugergyldighed']
+        )
+
+    def test_edit_employee(self):
```
cmoesgaard: Ditto.

backend/mora/mapping.py
```diff
@@ -66,10 +68,10 @@
 
 # Org unit
 ORG_UNIT_TYPE = 'org_unit_type'
-NAME = 'name'
+# NAME = 'name' - see under common
 PARENT = 'parent'
 ADDRESSES = 'addresses'
-LOCATION = 'location'
+# LOCATION = 'location' - see under common
```
danchr: Ditto.

backend/mora/service/employee.py
```diff
@@ -35,6 +35,200 @@
                             url_prefix='/service')
 
 
+class BrugerRequestHandler(handlers.RequestHandler):
+    __slots__ = ['details_requests']
```
danchr: I know this is a total nit, but `__slots__` is usually a tuple not a list. Although unlikely to matter in practice, tuples are unmodifiable:
```suggestion
    __slots__ = ('details_requests',)
```

backend/mora/service/employee.py
```diff
@@ -35,6 +35,200 @@
...
+    role_type = "employee"
+
+    def prepare_create(self, req):
+
```
danchr: Extraneous newline.

backend/mora/service/employee.py
```diff
@@ -35,6 +35,200 @@
...
+        """Create a new employee
```
danchr: Although it probably out to be, this isn't actually where we get the documentation. Rather, it's extracted from `detail_writing` — which you also provide. Any docstring here should be more about the code. At least for now :/

(Note that the other classes don't actually provide one.)

backend/mora/service/detail_writing.py
```diff
@@ -393,6 +393,47 @@ def edit():
...
+    The **original** and **data** objects follow the same structure.
+    Every field in **original** is required, whereas **data** only needs
+    to contain the fields that need to change along with the validity dates.
+
+    :<jsonarr string cpr_no: The cpr number of the employee
+    :<jsonarr string name: The Name of the employee
+    :<jsonarr object validity: The validities of the changes.
+
+    .. sourcecode:: json
+        [{
```
danchr: In order for Sphinx to pick up the directory, you need a newline between the directive and the text. Then, you'll get a warning about your JSON not being actually valid JSON. (`None` ≠ `null`; no final commas; missing quotes.)

To build the docs, use `./flask.sh docs`.

backend/tests/test_integration_employee.py
jgj-magenta-aps: ditto-ditto fixed

backend/mora/service/employee.py
```diff
@@ -35,6 +35,200 @@
                             url_prefix='/service')
 
 
+class BrugerRequestHandler(handlers.RequestHandler):
```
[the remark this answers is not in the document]
jgj-magenta-aps: Yes - and done

backend/mora/mapping.py
jgj-magenta-aps: removed are now really removed

backend/mora/service/employee.py
jgj-magenta-aps: yup

backend/mora/service/employee.py
jgj-magenta-aps: I had skipped this exercise - I only had the detail_writing-edit part documented.
I have now moved the create-docu there too and outfitted it with a "type": "employee"

PR #208, opened 2018-10-29
Updated to use flask_saml_sso for auth flow

cmoesgaard:
Depends on https://github.com/magenta-aps/flask_saml_sso/pull/1

https://redmine.magenta-aps.dk/issues/23928


danchr:
I think we should remove all vestiges of the preexisting authentication support as part of this PR 🙂

backend/mora/auth/base.py
```diff
@@ -14,45 +14,23 @@
...
-
-        if assertion:
-            r.headers['Authorization'] = assertion
-
-        return r
-
-
+@flask_saml_sso.requires_auth
 @blueprint.route('/user', methods=['GET'])
 def get_user():
```
danchr: Perhaps we should fold this into the module as well?

cmoesgaard: It's a thing that's only specific for OS2MO, and additionally requires configuration of which field in the attributes contains the name, so it isn't a thing that would work out-of-the-box anyway.

cmoesgaard:
> I think we should remove all vestiges of the preexisting authentication support as part of this PR slightly_smiling_face

I agree. I'll do that.

backend/mora/auth/base.py
cmoesgaard: But what we _could_ do, is make it easier to get the attributes from the session, so you don't have to remember that attributes are saved in `samlUserdata`. I'll do something about that.

cmoesgaard:
Yes, I realize that. This PR depends on the PR for `flask_saml_sso` being merged in, so a proper release can be made of that first. Whether that results in that package being pushed to PyPI or what, I don't know yet, but the integration to it will remain unchanged, so you can still review _that_ part of it, at least :smile:

backend/mora/app.py
```diff
@@ -12,13 +12,14 @@
...
-import flask_session
 
 from . import exceptions
 from . import service
 from . import settings
 from . import util
-from .auth import base, sso
+from .auth import base
+
+import flask_saml_sso
```
danchr: Technically, this is an external module like `flask` — even if written by us — so it should go into the same location as `flask_session` IMO.

backend/mora/cli/lora.py
```diff
@@ -11,25 +11,21 @@
...
 import pyexcel
-import requests
-import urllib3
 
 from .. import exceptions
 from .. import lora
 from .. import settings
-from ..auth import base as auth_base, tokens as auth_tokens
+
+import flask_saml_sso
```
danchr: Likewise.

backend/mora/cli/lora.py
```diff
@@ -40,55 +36,15 @@ def group():
...
 
 def requires_auth(func):
-    @click.option('--user', '-u',
-                  help="account user name")
-    @click.option('--password', '-p',
-                  help="account password")
-    @click.option('--insecure', '-k', is_flag=True,
-                  help="disable SSL/TLS security checks")
+    @click.option('--token', '-t',
+                  help="auth session token")
```
danchr:
```suggestion
                  help="API token")
```

backend/mora/cli/lora.py
```diff
@@ -40,55 +36,15 @@ def group():
...
     def wrapper(*args, **options):
-        insecure = options.pop('insecure')
-
-        if insecure:
-            warnings.simplefilter('ignore', urllib3.exceptions.HTTPWarning)
-
-            lora.session.verify = False
-        else:
-            warnings.simplefilter('error', urllib3.exceptions.HTTPWarning)
+        token = options['token']
```
danchr: I'd inline this :)

backend/mora/lora.py
```diff
@@ -11,18 +11,17 @@
 import collections
 import functools
 import itertools
-import uuid
-
 import requests
+import uuid
```
danchr: `uuid` is an internal module.

backend/mora/service/__init__.py
```diff
@@ -7,15 +7,20 @@
 #
 
 from . import address
+from . import association
```
danchr: I assume this will generate a lot of warnings by `flake8`? We should probably have a `nocov` or `noqa` comment, or perhaps include them in `__all__` if that silences it.

backend/mora/service/address.py
```diff
@@ -199,9 +199,9 @@
 
 import collections
 import json
-import re
 
 import flask
+import re
```
danchr: Part of standard library.

backend/mora/service/detail_writing.py
```diff
@@ -20,20 +20,10 @@
 
 '''
 
-import typing
-
 import flask
+import typing
```
danchr: Part of standard library.

docs/dev/auth.rst
```diff
@@ -23,31 +29,9 @@ the SP with correct configuration.
 Configuration
 """""""""""""
 
-The following configuration entries are relevant for SAML SSO:
+Configuration of SAML SSO is documented in the readme for Flask SAML SSO.
+
+The following additional configuration entries exist for auth in OS2MO.
```
danchr: How about roles?

backend/mora/settings.py
```diff
@@ -23,24 +23,24 @@
...
+SAML_AUTH_ENABLE = False
 SAML_IDP_METADATA_URL = '/url/to/sso/metadata'
 SAML_IDP_METADATA_FILE = None
+SAML_IDP_INSECURE = False
 SAML_USERNAME_ATTR = ''
 SAML_KEY_FILE = None
 SAML_CERT_FILE = None
 SAML_REQUESTS_SIGNED = False
 SAML_DUPLICATE_ATTRIBUTES = True
+SAML_API_TOKEN_RESTRICT = False
```
danchr: Are these used by `flask-saml-sso`?

cmoesgaard: Yes.

docs/dev/auth.rst
cmoesgaard: What do you mean?

danchr: I was thinking of group membership, etc., shouldn't we mention that in the documentation?

backend/requirements-test.txt
```diff
@@ -17,3 +16,15 @@ testing.postgresql>=1.3.0
 flask-shell-ipython
 python-dotenv
 watchdog
+
+# All requirements for oio_rest
+-e git+https://github.com/magenta-aps/mox@development#egg=oio-rest&subdirectory=oio_rest
+pytz==2018.5
+Jinja2==2.10
```
danchr: `mora` already depends on some of these packages — at the very least `Jinja2` & `Werkzeug` — and we should purge the dependencies to avoid duplicates.

backend/requirements-test.txt
```diff
@@ -17,3 +16,15 @@ testing.postgresql>=1.3.0
...
+MarkupSafe==1.0
+Werkzeug==0.14.1
+psycopg2cffi-compat==1.1
+python3-saml==1.4.1
```
danchr: Is `python3-saml` still a dependency of `oio_rest`?

docs/dev/auth.rst
cmoesgaard: Group membership in terms of being able to create API tokens etc? That's missing, yes. I'd rather add that to the docs for `flask_saml_sso` as that's where the code is.

backend/requirements-test.txt
cmoesgaard: No, it isn't. I'll clean these up.

PR #221, opened 2018-11-14
jenkins: set a timeout for the warnings scanner

danchr:

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -98,11 +98,6 @@ pipeline {
 
   post {
     always {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -120,6 +115,13 @@ pipeline {
       junit healthScaleFactor: 200.0,           \
         testResults: '*/build/reports/*.xml'
 
+      timeout (1) {
```
cmoesgaard: Is this a one second timeout? Is that enough in every situation?

danchr: It's one minute, which ought to be enough.

PR #227, opened 2018-11-28
Feature/25601-linting

andersjepsen:
just ran yarn lint to fix most linting issues. A few needed human intervention to fix, most notably MoManagerEntry which were having 'unexpected side effects' as the linter said.


frontend/src/components/ButtonSubmit.vue
```diff
@@ -14,31 +14,31 @@
 </template>
 
 <script>
-  /**
+/**
    * A button submit component.
```
n-singh: It seems you've forgot to insertion the comments. It's the same issue in all the components.
```suggestion
/**
  * A button submit component.
  */
```

PR #235, opened 2018-12-13
Prevent edits in the past.

cmoesgaard:
This adds validation in front- and backend to ensure edits are not performed on dates before today's date.
Note, that this is not implemented for addresses, pending the rewrite and remodeling of these as discrete objects.

https://redmine.magenta-aps.dk/issues/25394

- ~Have you updated the documentation?~

### DIFF
--- a/backend/mora/validator.py
+++ b/backend/mora/validator.py
@@ -312,3 +312,15 @@ def does_employee_have_active_engagement(employee_uuid, valid_from, valid_to):
 
     if not valid_effects:
         exceptions.ErrorCodes.V_NO_ACTIVE_ENGAGEMENT(employee=employee_uuid)
+
+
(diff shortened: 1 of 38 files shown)


backend/mora/validator.py
```diff
@@ -312,3 +312,15 @@ def does_employee_have_active_engagement(employee_uuid, valid_from, valid_to):
 
     if not valid_effects:
         exceptions.ErrorCodes.V_NO_ACTIVE_ENGAGEMENT(employee=employee_uuid)
+
+
+def is_edit_from_date_before_today(from_date: datetime.datetime):
+    """Check if a given edit date is before today. If so, raise exception"""
+    today = datetime.datetime.combine(
+        datetime.date.today(),
```
danchr: We have some logic for handling dates in `util.py`:
```python
today = util.now().date()
```

PR #248, opened 2019-01-21
25881: Removed db_structure patching

cmoesgaard:
:man_shrugging:


danchr:
You need to touch `mora/cli/base.py` as well.

PR #255, opened 2019-01-25
Feature/25053 function addresses

cmoesgaard:
https://redmine.magenta-aps.dk/issues/25053
https://redmine.magenta-aps.dk/issues/25409

Review: https://redmine.magenta-aps.dk/issues/26659


backend/mora/service/addresses/base.py
```diff
@@ -0,0 +1,92 @@
...
+    def get_lora_properties(self):
+        return []
+
+    def get_lora_address(self):
+        return {
+            'objekttype': self.scope,
+            'urn': self.get_urn(),
+        }
+
+    def get_mo_properties(self):
```
danchr: Add a docstring; what is properties; why is it different from the address below; why are they not just one?

Also, the `_mo_` chunk is a new naming convention that does not occur anywhere else in the codebase. I think the closest is `_obj` to signify that something is, in fact, an object. However, since this is a conversion method rather than a simple accessor, an `as_` prefix might be clearer. But that's also new. If you convert the accessors to properties (or remove them), the `get_` would suffice.

backend/mora/service/addresses/dar.py
```diff
@@ -0,0 +1,121 @@
+#
+# Copyright (c) 2017-2018, Magenta ApS
+#
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+import requests
+import flask
+from . import base
```
danchr: Insert a newline between the 3rd party modules and our own. And perhaps sort them alphabetically?
```suggestion

from . import base
```

backend/mora/service/addresses/base.py
```diff
@@ -0,0 +1,92 @@
...
+        value = urn[len(cls.prefix):]
+        return cls(value)
+
+    @classmethod
+    def from_request(cls, request):
+        """Initialize handler from MO object"""
+        value = util.checked_get(request, mapping.VALUE, "", required=True)
+        return cls(value)
+
+    def get_urn(self):
```
danchr:
```suggestion
    @property
    def urn(self):
```

backend/mora/service/addresses/base.py
```diff
@@ -0,0 +1,92 @@
...
+        return self.prefix + self.value
+
+    def get_value(self):
```
danchr:
```suggestion
    @property
    @abstractmethod
    def value(self):
```
1) `get_*` accessors are poor form in Python; that's what `@property` is for.
2) Having a `get_value` method that does nothing other than return `self.value` is entirely redundant; why not simply assign to — and access — `value` instead?

backend/mora/service/association.py
```diff
@@ -77,9 +80,10 @@ def prepare_create(self, req):
             funktionstype=association_type_uuid,
             opgaver=[
                 {'uuid': job_function_uuid}] if job_function_uuid else None,
-            adresser=[
-                address.get_relation_for(address_obj),
-            ] if address_obj else None,
+            adresser=[{
+                'uuid': addr_id,
+                'objekttype': 'organisationfunktion'
```
danchr: This feels vaguely iffy… We seem to be using the `adresser` relation for something rather unrelated, i.e. pointing to an `organisationfunktion`. IIRC we don't do that for IT systems, which went through a similar iteration.

backend/mora/service/association.py
```diff
@@ -152,13 +159,14 @@ def prepare_edit(self, req: dict):
...
-                data.get(mapping.ADDRESS) or original_data[mapping.ADDRESS]
-            )
+            address_obj = util.checked_get(data, mapping.ADDRESS, {})
 
             update_fields.append((
                 mapping.SINGLE_ADDRESS_FIELD,
-                address.get_relation_for(address_obj),
+                {
+                    'objekttype': 'organisationfunktion',
+                    'uuid': address_obj.get(mapping.UUID),
```
danchr: See above.

backend/mora/service/detail_reading.py
```diff
@@ -439,12 +389,11 @@ def get_detail(type, id, function):
     search = {
         info.search: id,
     }
-    scope = getattr(c, info.scope)
```
danchr: Is this attribute now dead?

backend/mora/service/detail_reading.py
```diff
@@ -500,6 +444,20 @@ def get_user_key(effect):
     # return them as a list rather than a single object; otherwise, we
     # just return the first hit
     converters = {
+        'address': {
```
danchr: If dead, euthanise without mercy.

backend/mora/service/detail_reading.py
```diff
@@ -642,11 +599,45 @@ def as_values(vs):
                     cache[v] = None
 
     # fetch and convert each object once, rather than multiple times
+    #
+
+    # handle cross-function links first, the only instance being this
+    # detail referring to an address
+    address_functions = collections.OrderedDict(
```
danchr: I'd write this further up if I could; the support for handling string value in `as_values()` is now unused dead, and may safely be removed.

backend/mora/service/addresses/base.py
cmoesgaard: 1. I'll see what I can do about adding `@property`
2. Because there are cases in the specific handler where `get_value()` _doesn't_ just return the value.

backend/mora/service/addresses/base.py
cmoesgaard: The naming was attempt to be more clear in the naming of the objects we're working with, which we haven't really made a clear distinction between so far, IMO.

We basically operate with two different groups of models. The ones we use towards LoRa, and the ones we use towards the UI.  Calling it `get_properties` wouldn't make it super obvious, that this is actually meant for the specific model used towards the UI.

backend/mora/service/address.py
```diff
@@ -5,197 +5,6 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 #
-'''Addresses
```
danchr: Is there any particular reason for moving this out of the code?

backend/mora/service/addresses/base.py
danchr: > The naming was attempt to be more clear in the naming of the objects we're working with, which we haven't really made a clear distinction between so far, IMO.

That makes sense — I'd be inclined towards something like `_api_` but `_mo_` is fine too 🙂

> We basically operate with two different groups of models. The ones we use towards LoRa, and the ones we use towards the UI. Calling it `get_properties` wouldn't make it super obvious, that this is actually meant for the specific model used towards the UI.

AFAICT you can quite safely fold `get_mo_properties()` and `get_mo_address()` into one; the only caller expands _both_ into the same dictionary, right?

backend/mora/service/addresses/dar.py
```diff
@@ -0,0 +1,121 @@
...
+    scope = 'DAR'
+    prefix = 'urn:dar:'
+
+    def __init__(self, value):
+        super().__init__(value)
+
+        try:
+            self.address_object = self._fetch_from_dar(value)
+        except Exception as exc:
+            self.address_object = {
```
danchr: We can do better than this, now!

Previously, the only useful data we could return when offline, was some text explaining that we couldn't, in fact, return anything useful. Now, we have an entire object where we can stash all sorts of useful data, such as a stringification off the actual address. If we do so, we might even be able to avoid the step of looking up in DAR when reading, although that's not of great importance. But we could simplify the reading code: the only thing DAWA's `historik` endpoints offer us is a textual description, but if we already have that, why query it?

backend/mora/service/association.py
cmoesgaard: I based it on your previous code, but I see your point :smile: 
I'll change it.

backend/mora/service/detail_reading.py
danchr: > Yes. It wasn't really used for anything before.

🗡 ⚔️ 💀

backend/mora/service/addresses/phone.py
```diff
@@ -0,0 +1,72 @@
...
+
+    @classmethod
+    def from_request(cls, request):
+        value = util.checked_get(request, mapping.VALUE, "", required=True)
+        visibility = util.get_mapping_uuid(request, mapping.VISIBILITY)
+
+        return cls(value, visibility)
+
+    def get_value(self):
+        return self.value[3:]
```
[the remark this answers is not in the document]
cmoesgaard: Value, is meant as the _editable_ value - e.g. just the eight digits of the phone number without +45. Pretty value (or name, as I just now renamed it to) is meant to be the user-facing prettyfied value.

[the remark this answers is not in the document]
cmoesgaard: I added docstrings to clear this up.

danchr:
One minor thing that rubs me the wrong way: we have a module, `service.address` and a package, `service.addresses` with almost exactly the same name, apart from grammar. Couldn't we fold them into one? Perhaps have `service.address` ➡️  `service.address.__init__` and the rest as modules therein?

backend/tests/test_service_address.py
```diff
@@ -319,50 +305,14 @@ def test_many_addresses(self, m):
...
-                lora.Connector(),
-                {
-                    'objekttype': 'kaflaflibob',
-                    'uuid': '00000000-0000-0000-0000-000000000000',
-                },
-            )
-
-    @util.mock()
-    @freezegun.freeze_time('2010-01-01', tz_offset=1)
-    def test_bad_phone(self, m):
```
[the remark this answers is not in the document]
cmoesgaard: I'm not really sure what this is actually testing.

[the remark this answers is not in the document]
danchr: > I'm not really sure what this is actually testing.

The test consists of reading a value from an address that disagrees with the class; e.g. what happens when an address of type `EMAIL` has a phone number stored? Yeah, it's rather esoteric, but a useful thing to test.

backend/mora/service/addresses/dar.py
cmoesgaard: I feel like these improvements, while seeming like very good ideas, are kind of out of scope of this feature. As of now, I've done my best to reimplement the existing functionality, which appears to be working. I suggest we implement these ideas at a later date.

backend/mora/service/association.py
```diff
@@ -33,32 +33,35 @@ class AssociationRequestHandler(handlers.OrgFunkRequestHandler):
...
 
-        org_uuid = org_unit['relationer']['tilhoerer'][0]['uuid']
+        org_uuid = org_unit_obj['relationer']['tilhoerer'][0]['uuid']
         job_function_uuid = util.get_mapping_uuid(req, mapping.JOB_FUNCTION)
         association_type_uuid = util.get_mapping_uuid(req,
                                                       mapping.ASSOCIATION_TYPE,
                                                       required=True)
-        address_obj = util.checked_get(req, mapping.ADDRESS, {})
+
+        addr_id = util.get_mapping_uuid(req, mapping.ADDRESS)
```
danchr:
```suggestion
        function_id = util.get_mapping_uuid(req, mapping.ADDRESS)
```
DAR addresses have another UUID, so `addr_id` might be confusing.

backend/mora/service/address_handler/base.py
```diff
@@ -0,0 +1,148 @@
...
+    def __new__(mcls, name, bases, namespace):
+        cls = super().__new__(mcls, name, bases, namespace)
+
+        cls._register()
+
+        return cls
+
+
+class AddressHandler(metaclass=_AddressHandlerMeta):
+    __slots__ = 'scope', 'prefix', '_value'
```
danchr: Note that this adds very little unless all the subclasses also have a `__slots__` declaration.

backend/mora/service/address_handler/dar.py
```diff
@@ -0,0 +1,127 @@
...
+            else None
+        )
+
+        return href
+
+    def get_mo_address_and_properties(self):
+        if self.address_object.get(mapping.ERROR):
+            return self.address_object
+        else:
+            return super().get_mo_address_and_properties()
```
danchr: This all feels a bit too magical; optionally calling `super()`…

backend/tests/test_service_address.py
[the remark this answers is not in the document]
cmoesgaard: It's esoteric, yes, and I'm unsure of when it will actually be relevant. The addresses are created based on the scopes, so I can't see how the objects can suddenly become inconsistent.

backend/mora/service/address_handler/dar.py
cmoesgaard: For the most part, this is to mimic the original implementation, returning an error object rather than the address object in case an error occurred.

danchr: > For the most part, this is to mimic the original implementation, returning an error object rather than the address object in case an error occurred.

Ah, okay. Perhaps stashing them on separate fields, then?

```python
        if self.error is not None:
            return self.error
        else:
            return super().get_mo_address_and_properties()
```

…or just a slightly cleaner an more readable version of the original.

```python
        if mapping.ERROR in self.address_object:
            return self.address_object
        else:
            return super().get_mo_address_and_properties()
```

backend/mora/service/address_handler/base.py
cmoesgaard: I use it simply to declare the properties available on the class. What's the alternative, declaring `scope = None` etc. for every property?

danchr: > I use it simply to declare the properties available on the class. What's the alternative, declaring `scope = None` etc. for every property?

That's more explicit, actually, given that `__slots__` does not assign a default value. Given:

```python
class X:
    __slots__ = ('x', 'y')

    def __init__(self):
        print(self.x)

X()
```
The result is:

```python-traceback
Traceback (most recent call last):
  File "<stdin>", line 7, in <module>
  File "<stdin>", line 5, in __init__
AttributeError: x
```

backend/mora/service/address_handler/dar.py
cmoesgaard: I'll save it on a separate field.

backend/mora/service/association.py
cmoesgaard: I've renamed it to `addr_function_id`

backend/mora/service/addresses/dar.py
```diff
@@ -0,0 +1,121 @@
...
+from ... import mapping
+
+session = requests.Session()
+
+NOT_FOUND = "Ukendt"
+
+
+class DARAddressHandler(base.AddressHandler):
+    scope = 'DAR'
+    prefix = 'urn:dar:'
```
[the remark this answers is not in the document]
cmoesgaard: It would be nice, but it seems out of scope.

backend/mora/service/addresses/dar.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def get_mo_address(self):
+        if self.address_object.get(mapping.ERROR):
+            return self.address_object
+        else:
+            return super().get_mo_address()
+
+    @staticmethod
+    def _fetch_from_dar(addrid):
+        for addrtype in (
```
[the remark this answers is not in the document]
cmoesgaard: It seems out of scope for now, but we should keep it in mind for an eventual improvement in the future.

PR #265, opened 2019-02-06
Bugfix/xxxxx more fixes

cmoesgaard:

### DIFF
--- a/backend/mora/service/address_handler/phone.py
+++ b/backend/mora/service/address_handler/phone.py
@@ -8,6 +8,8 @@
 import re
 
 from . import base
+from .. import facet
+from ... import lora
 from ... import mapping
 from ... import util
 
@@ -66,9 +68,8 @@ class PhoneAddressHandler(base.AddressHandler):
     def _get_mo_properties(self):
         properties = super()._get_mo_properties()
         if self.visibility:
+            c = lora.Connector()
             properties.update({
-                mapping.VISIBILITY: {
-                    'uuid': self.visibility
-                }
+                mapping.VISIBILITY: facet.get_one_class(c, self.visibility)
             })
         return properties
--- a/backend/mora/service/detail_reading.py
+++ b/backend/mora/service/detail_reading.py
@@ -647,6 +647,13 @@ def get_detail(type, id, function):
         for classid, classobj in c.klasse.get_all(uuid=class_cache)
     })
 
+    def is_empty_lora_object(obj):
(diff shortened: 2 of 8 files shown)


backend/mora/service/detail_reading.py
```diff
@@ -647,6 +647,13 @@ def as_values(vs):
         for classid, classobj in c.klasse.get_all(uuid=class_cache)
     })
 
+    def is_empty_lora_object(obj):
+        empty = not (obj.get('relationer') or
```
danchr: Don't use `get()` when you mean `in`? Alternatively:

```python
return not obj.keys() & {'relationer', 'attributter', 'tilstande'}
```

backend/mora/service/detail_reading.py
```diff
@@ -655,6 +662,7 @@ def as_values(vs):
             **address.get_one_address(funcobj),
         }
         for funcid, funcobj in address_functions.items()
+        if not is_empty_lora_object(funcobj)
```
danchr: Perhaps:

```python
if util.is_reg_valid(funcobj)
```

I believe this is consistent with the other reading code and arguably a more thorough check.

backend/mora/service/address_handler/phone.py
```diff
@@ -66,9 +68,8 @@ def get_lora_properties(self):
     def _get_mo_properties(self):
         properties = super()._get_mo_properties()
         if self.visibility:
+            c = lora.Connector()
             properties.update({
-                mapping.VISIBILITY: {
-                    'uuid': self.visibility
-                }
+                mapping.VISIBILITY: facet.get_one_class(c, self.visibility)
```
danchr: Should we cache these?

cmoesgaard: Caching classes is some of the most low hanging fruit as they bound to almost never change, so we should definitely do that, but I am not really sure how to do it yet.

I'll add a ticket for it, (if there isn't one already) and then we can talk about the design for it at some point.

backend/tests/test_address_handlers.py
```diff
@@ -336,7 +339,11 @@ def test_get_mo_address(self):
         }
 
         # Act
-        actual = address_handler.get_mo_address_and_properties()
+        with patch(
+            'mora.service.address_handler.phone.facet.get_one_class',
```
danchr: Oy vey, that's a long name 😁  Why not do a proper module-level import and just patch `facet.get_one_class`?

PR #285, opened 2019-02-19
Feature/26886 - addresses duplicate and employee move many fix

n-singh:
~- [ ] Have you updated the documentation?~
~- [ ] Have you written tests for your changes?~


frontend/src/views/organisation/workflows/MoOrganisationUnitCreate.vue
```diff
@@ -115,14 +115,16 @@ export default {
...
-        this.addresses.push(this.postAddress, this.phone)
-        this.addresses.forEach(a => {
-          if (!a.validity) {
-            a.validity = this.entry.validity
-          }
-          a.org = this.$store.getters['organisation/GET_ORGANISATION']
-        })
-        this.entry.details = this.addresses
+        if (!vm.backendValidationError) {
+          this.addresses.push(this.postAddress, this.phone)
```
danchr: From my understanding, `backendValidationError` gets set whenever a request to the backend fails. Once this occurs, you will no longer store that new addresses; correct?

It just so happens that this may work due to the exact instances of `this.postAddress` and `this.phone` appearing in the array, and Vue merely updates those. That being said, I think it would be cleaner to simply reassign `this.addresses`, or explicitly clearing it prior to the assignment.

frontend/src/views/organisation/workflows/MoOrganisationUnitCreate.vue
```diff
@@ -115,14 +115,14 @@ export default {
         let vm = this
         this.isLoading = true
 
-        this.addresses.push(this.postAddress, this.phone)
-        this.addresses.forEach(a => {
+        let details = [this.postAddress, this.phone, ...this.addresses]
+        details.forEach(a => {
           if (!a.validity) {
```
danchr: We potentially still have the issue here, don't we? One option is to use `map` instead of `forEach`; by allocating new objects, you won't leave polluted state behind.

frontend/src/views/organisation/workflows/MoOrganisationUnitCreate.vue
```diff
@@ -115,14 +115,14 @@ export default {
         let vm = this
         this.isLoading = true
 
-        this.addresses.push(this.postAddress, this.phone)
-        this.addresses.forEach(a => {
+        let details = [this.postAddress, this.phone, ...this.addresses]
+        details.foreach(a => {
```
danchr:
```suggestion
        details.forEach(a => {
```

PR #286, opened 2019-02-20
24474: Fixed incorrect handling of name property in DAR address handl…

cmoesgaard:
…er when lookup failed, causing detail creation to raise exception

### DIFF
--- a/backend/tests/test_address_handlers.py
+++ b/backend/tests/test_address_handlers.py
@@ -101,6 +101,36 @@ class DarAddressHandlerTests(util.TestCase):
         # Assert
         self.assertEqual(expected, actual)
 
+    def test_failed_lookup(self, mock):
+        """Ensure that failed lookups are handled appropriately"""
+        # Arrange
+
+        # Nonexisting DAR UUID should fail
+        value = '300f16fd-fb60-4fec-8a2a-8d391e86bf3f'
+        address_handler = self.handler(value)
+
+        expected_name = dar.NOT_FOUND
+        expected_href = None
+        expected_mo_address = {
+            'error': 'No mock address: GET '
+                     'https://dawa.aws.dk/adresser?'
+                     'id=300f16fd-fb60-4fec-8a2a-8d391e86bf3f'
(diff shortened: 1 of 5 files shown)


backend/tests/test_address_handlers.py
```diff
@@ -101,6 +101,36 @@ def test_get_lora_properties(self, mock):
...
+            'value': '300f16fd-fb60-4fec-8a2a-8d391e86bf3f'
+        }
+
+        # Act
+        actual_name = address_handler.name
+        actual_href = address_handler.href
+        actual_mo_address = address_handler.get_mo_address_and_properties()
+
+        # Assert
+        self.assertEqual(expected_name, actual_name)
```
danchr: Why three asserts and not just one comparing the full value?

backend/tests/test_integration_address.py
```diff
@@ -1401,23 +1439,24 @@ def test_missing_address(self, mock):
         self.assertRequestResponse(
             '/service/ou/{}/details/address'.format(unitid),
             [{
-                'address_type': address_class,
-                'name': 'Ukendt',
-                'value': addrid,
-                'error': 'no such address {!r}'.format(addrid),
+                'address_type': {
```
danchr: Erm, the name of this test is `test_missing_address`, yet it now succeeds. Does it still test a properly missing address?

cmoesgaard: It succeeds, yes, in the same way that the previous one did. We're unable to read the address from DAR but still return _something_ to be rendered in the UI.

danchr: > It succeeds, yes, in the same way that the previous one did. We're unable to read the address from DAR but still return _something_ to be rendered in the UI.

Yeah, I saw that afterwards; I was a bit confused by it appearing slightly less fail-y.

cmoesgaard: I omitted the `error` part of the response. As far as I could tell, we didn't use it for anything. We log the error as well, so we should be good :slightly_smiling_face:

PR #287, opened 2019-02-21
tests: remove patching that belongs in LoRA upstream

danchr:
We have `test_support` now; anything that touches LoRA internals
belongs there, so that we can modify it exclusively upstream.

- [x] ~Have you updated the documentation?~ N/A
- [x] ~Have you performed a smoke test of the application?~ N/A
- [x] ~Have you written tests for your changes?~ N/A
- [x] ~Have you updated the release notes?~ N/A

### DIFF
--- a/backend/tests/util.py
+++ b/backend/tests/util.py
@@ -259,14 +259,16 @@ def load_sample_structures(*, verbose=False, minimal=False, check=False,
 
 @contextlib.contextmanager
 def override_settings(**overrides):
-    orig_settings = {k: getattr(settings, k) for k in overrides}
-    settings.__dict__.update(overrides)
-    yield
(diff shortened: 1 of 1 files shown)


backend/tests/util.py
```diff
@@ -543,10 +543,6 @@ def setUp(self):
                 self.lora_port,
             )),
             patch('oio_rest.app.settings.LOG_AMQP_SERVER', None),
```
cmoesgaard: Shouldn't this line be removed/replaced as well?

danchr: > Shouldn't this line be removed/replaced as well?

Good point; although it's a user-facing setting, the _location_ of that variable isn't. (We might move `settings` in the future, as we have in the past.) I'll fix it!

PR #296, opened 2019-02-28
Remove association address

agger-magenta:


frontend/src/components/MoEntry/MoAssociationEntry.vue
```diff
@@ -8,6 +8,11 @@
         required
       />
 
+      <mo-address-picker
```
cmoesgaard: This should be removed as well.

PR #302, opened 2019-03-05
PROD_MODE inverses to DUMMY_MODE

jgj-magenta-aps:
No - there currently is no configuration for the configuration.
- [X] Have you performed a smoke test of the application?
- [X] Have you written tests for your changes?
- [X] Have you updated the release notes?


backend/mora/integrations/serviceplatformen.py
```diff
@@ -9,18 +9,52 @@
...
+import logging
+import pathlib
+import requests
 
 from .. import util
 from .. import settings
+from .. import exceptions
+
+
+logger = logging.getLogger("flask.app")
```
cmoesgaard: I don't think `flask.app` is a good name for a logger here. Use `__name__`?
Also, we could use the built-in logger in flask:
`flask.current_app.logger.getChild(__name__)`

backend/flask.sh
```diff
@@ -18,4 +18,7 @@ fi
 export FLASK_ENV=development
 export FLASK_APP=mora.app:app
 
+# change default setting during test
+[ "$1" = "test" ] && export OS2MO_DUMMY_MODE=true
```
cmoesgaard: This is not an ideal place to do this. It overrides the default setting but _only_ when run through `flask.sh`. Consider instead overriding it in the test class in `tests/util.py`

backend/mora/util.py
```diff
@@ -73,6 +73,8 @@
     2 * 60**2: DEFAULT_TIMEZONE,
 }
 
+logger = logging.getLogger(__name__)
```
danchr: I'd just inline this one; you'll only be using it once, and ideally, we should use the Flask app logger instead. (Which might not work in this case…)

backend/tests/test_app_check_config.py
```diff
@@ -0,0 +1,59 @@
+#
```
danchr: I'd fold this file into `test_cpr.py` — they seem quite related, and both are rather small anyway…

backend/mora/util.py
```diff
@@ -283,19 +285,26 @@ def update_config(mapping, config_path, allow_environment=True):
...
+            k[6:]: v
             for k, v in os.environ.items()
             if k.startswith('OS2MO_')
         }
 
         for key in overrides.keys():
-            print(
-                ' * Using override OS2MO_{}={!r}'.format(key, overrides[key]),
-                file=sys.stderr)
+            if overrides[key] in ['True', 'true']:
```
danchr: Utility function? Also, we have another function which does _almost_ the same; `get_args_flag()` — perhaps they should share the logic?

(`TRUE`, `1`, and `y` are common ways to opt into something. `fALsE` is just being obnoxious 😉)

backend/tests/test_app_check_config.py
jgj-magenta-aps: I agree!, DUMMY_MODE is a bad name, but general: It relays the information that all integrations are turned off, which is what I think is the meaning: The inverse of PROD_MODE (where all integrations are ago) - we just have only one at present.

Could We agree on another name: DEMO_MODE / STUB_MODE

I will put in checks for the 4 uuids as well and move the tests into test_cpr.

backend/mora/util.py
jgj-magenta-aps: We are not yet in a flask context, so flask is a no go.
Inline?
Do You mean logging.getLogger(__name__).warn(...)?

backend/mora/util.py
jgj-magenta-aps: I agree somewhat, but I also think what i did here was maybe wrong.:
A config file is different - it may relay settings to other modules that actually need 1, 'true' or y. 
-
Hence I may even back off from this and move it into 'check_config', so it is the checker, that does the conversion (only converts 'false' to False in it's own scope)
-
What do You think about that?

danchr:
> I'm not a fan of having a `DUMMY_MODE` setting; at the very least, the name should be more descriptive of its purpose and effect; something like `FAKE_SP_CPR`.

I forgot we actually agreed on `DUMMY_MODE`, so please feel free to disregard this one! :)

backend/.jenkins/common.sh
```diff
@@ -2,6 +2,8 @@
 
 set -ex
 
+export OS2MO_DUMMY_MODE=true
```
danchr: You shouldn't need to set that here; the tests should assign it themselves as needed. After all, developers frequently run the tests outside Jenkins 🙂

jgj-magenta-aps: Yes - it is a remnant. Sorry. It is set in tests/util.py also.

backend/mora/integrations/serviceplatformen.py
jgj-magenta-aps: I use name - we have no flask app at that time

backend/mora/util.py
```diff
@@ -287,15 +287,23 @@ def update_config(mapping, config_path, allow_environment=True):
...
 
         for key in overrides.keys():
+            if overrides[key] in ['True', 'true']:
+                overrides[key] = True
+            if overrides[key] in ['false', 'False']:
+                overrides[key] = False
             print(
-                ' * Using override OS2MO_{}={!r}'.format(key, overrides[key]),
-                file=sys.stderr)
+                ' * Using configuration override {}={!r}'.format(
```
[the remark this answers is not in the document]
jgj-magenta-aps: logger is used now.

backend/.jenkins/common.sh
danchr: > Yes - it is a remnant. Sorry. It is set in tests/util.py also.

Could you please remove it, then? 🙂

jgj-magenta-aps: done - will be exciting to see how Jenkins copes.

PR #312, opened 2019-03-12
27687: Added support for specifying user_key on the various org funk …

cmoesgaard:
…relation payloads.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -1,6 +1,11 @@
 Version 0.16.0, in development
 ==============================
 
+New features
+------------
+
+* #27687: The various org funk relations now accept a ``user_key`` parameter.
+
 Bug fixes
 ---------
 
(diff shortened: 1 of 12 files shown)


NEWS.rst
```diff
@@ -1,3 +1,12 @@
+In development
+==============
+
+Internal changes
+----------------
+
+* #27687: The various org funk relations now accept a ``user_key`` parameter.
```
danchr: This should be a new feature, shouldn't it?

backend/mora/service/association.py
```diff
@@ -13,6 +13,7 @@
 This section describes how to interact with employee associations.
 
 '''
+import uuid
```
danchr: Unused?
```suggestion
```

NEWS.rst
cmoesgaard: I guess it is, yes

PR #353, opened 2019-05-21
Feature/29019 tak sinne

jgj-magenta-aps:
https://redmine.magenta-aps.dk/issues/29019

- [ ] This is a trivial change (and the rest of the list can be ignored)

**Please ensure the following is true:**

- [x] The application can be built and installed without errors
- [x] The feature/bugfix has been tested manually (if manually testable)
- [x] Tests have been written/updated for my change
- [ ] ~The documentation has been updated~
    - [ ] The documentation builds without warnings or errors
- [x] Release notes have been updated
- [x] The corresponding Redmine ticket has been set to `I test`

**If applicable:**

- [x] Changes have been made to the UI
    - [x] Screenshots of relevant changes to UI added to PR
- [ ] The API/data model has been changed/expanded
    - [ ] Redmine tickets have been created to update dependent systems
- [ ] The feature introduces changes relevant to the deployment process
    - [ ] Redmine tickets have been created to update deployment process
- [ ] This change affects the development workflow.
    - [ ] I have notified other developers by email or chat.


backend/mora/service/manager.py
```diff
@@ -51,10 +51,15 @@ def finder(cls, c, type, objid):
...
-                if not ou or mapping.PARENT not in ou:
+                if not ou:
                     return found
-                upperid = ou[mapping.PARENT][mapping.UUID]
-                found = super().finder(c, type, upperid)
+                upper = ou.get(mapping.PARENT)
+                if not upper:
+                    return found
+                objid = upper.get(mapping.UUID)
+                if not objid:
```
cmoesgaard: So, if I'm reading this right, this is in case we found a parent, but the parent doesn't have a UUID. Has that ever happened? Can it even happen?

jgj-magenta-aps: I have no idea. I read a value from a dictionary and with this I check not only if there was no UUID in the parent, but also if it was None. 

Same construct as with upper where it was indeed needed as there was a parent-key but the value was None.

The way I see it, it will hopefully never be needed, but it doesn't hurt.

NEWS.rst
```diff
@@ -27,6 +27,7 @@ New features
 Bug fixes
 ---------
 
+* #29019: Never ending loop in manager inheritance
```
cmoesgaard: Add this to 0.18.0 instead.

jgj-magenta-aps: Moved.

jgj-magenta-aps: Managers who are inherited are now star marked in ui
screenshot-star

frontend/src/store/modules/organisationUnit.js
```diff
@@ -13,12 +13,24 @@ const defaultState = () => {
     location: undefined,
     user_settings: {},
     details: {},
-    isLoading: false
+    isLoading: false,
+    ShowIfInherited: (org_unit_uuid, content) => {
```
cmoesgaard: I don't think this function belongs on the state.

PR #381, opened 2019-08-05
Send proper deletion request to fix validation error.

agger-magenta:
https://redmine.magenta-aps.dk/issues/22316

- [ ] This is a trivial change (and the rest of the list can be ignored)

**Please ensure the following is true:**

- [ ] The application can be built and installed without errors
- [ ] The feature/bugfix has been tested manually (if manually testable)
- [ ] Tests have been written/updated for my change
- [ ] The documentation has been updated
    - [ ] The documentation builds without warnings or errors
- [ ] Release notes have been updated
- [ ] The corresponding Redmine ticket has been set to `Needs review`
    - [ ] The issue has a link to this PR

**If applicable:**

- [ ] Changes have been made to the UI
    - [ ] Screenshots of relevant changes to UI added to PR
- [ ] The API/data model has been changed/expanded
    - [ ] Redmine tickets have been created to update dependent systems
- [ ] The feature introduces changes relevant to the deployment process
    - [ ] Redmine tickets have been created to update deployment process
- [ ] This change affects the development workflow.
    - [ ] I have notified other developers by email or chat.

### DIFF
--- a/backend/mora/common.py
+++ b/backend/mora/common.py
@@ -154,12 +154,6 @@ def update_payload(
         if field_tuple.type == mapping.FieldTypes.ADAPTED_ZERO_TO_MANY:
             # 'Fake' zero-to-one relation. Merge into existing list.
             updated_props = _merge_obj_effects(props, vals)
-            for p in updated_props:
-                if len(p) == 1 and 'virkning' in p:
-                    # Change periods with no UUID or URN to correct
-                    # link-changing syntax.
-                    p['uuid'] = ''
-                    p['urn'] = ''
         elif field_tuple.type == mapping.FieldTypes.ZERO_TO_MANY:
             # Actual zero-to-many relation. Just append.
             updated_props = vals
@@ -169,6 +163,13 @@ def update_payload(
             assert 0 <= len(vals) <= 1
             updated_props = vals
 
+        for p in updated_props:
+            # TODO: Fix this when the underlying bug in LoRa is resolved
+            # https://redmine.magenta-aps.dk/issues/31576
+            if (len(p) == 1 and 'virkning' in p and 'relationer' in
+                    field_tuple.path[0]):
+                p['uuid'] = ''
+                p['urn'] = ''
         payload = util.set_obj_value(payload, field_tuple.path, updated_props)
 
     return payload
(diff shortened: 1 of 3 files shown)


backend/mora/common.py
```diff
@@ -154,6 +154,12 @@ def update_payload(
         if field_tuple.type == mapping.FieldTypes.ADAPTED_ZERO_TO_MANY:
             # 'Fake' zero-to-one relation. Merge into existing list.
             updated_props = _merge_obj_effects(props, vals)
+            for p in updated_props:
```
cmoesgaard: This only handles the "empty" payloads for `ADAPTED_ZERO_TO_MANY` changes. We never want empty payloads, right? I think it would make sense to move this outside of the if/else structure, so we perform this step on _every_ update payload we create, regardless of field type.

agger-magenta:
> This change modifies some of the deepest payload-merge-magic in MO, so I don't think it qualifies as being trivial. 

OK, sounds reasonable.

>This function has some unit tests already, so you should add another one for this case as well, and make sure you've performed some smoke testing for the feature, added to the release notes, etc.

Yes, will do!

backend/mora/common.py
agger-magenta: This change is specific to the handling of relations. I don't think we have the same issue with attributes or states.

Byt you're right, I should check.

cmoesgaard: Field type in this case is a bit abstract, and refers to the "types" of cardinalities we work with in MO, towards LoRa - so the different branches in the function (zero-to-one, zero-to-many, adapted-zero-to-many) all potentially refer to relations.

agger-magenta: So this function _only_ applies to relations?

I tried setting a break point in the other branches, and the test suite never tries to send empty payloads in those, only for the adapted-zero-to-many. (Maybe these are handled differently by MO.)

But if this function indeed only applies to relations, it will be both safe and correct to replace all empty payloads with ones with uuid = urn   ''.

cmoesgaard: It applies to attributes and states as well, but we don't want those to be empty as well either, right? Actually, after having thought about it, maybe this fix belongs somewhere else, so we don't accidentally set `uuid` and `urn` in case we happen to be working with a "tilstand" where we forgot to set `gyldighed`.

agger-magenta: I will look into this, but the case of "empty" periods seem only to ever occur for relations.

cmoesgaard: That's true, but it doesn't change the fact that this function is meant to work on attributes and states, as well as relations, so _if_ we ever happen to have "empty" periods for any of those, we'll have another problem on our hands.

